PR #26, opened 2017-01-09
Introduce LifecycleObservable

hzsweers:
This is more of a proposal/exploration. We want to make the Observable behavior explicit for LifecycleScopeProvider, so this introduces `LifecycleObservable`. It's a thin shim just there to have its behavior documented in the type system and forces the implementer to think about its behavior.

Some things we might want to add:
- BehaviorSubject's `getValue()` API
- Verification that it is indeed behaving the way we expect

In use, it's easy to convert to it:

```java
myObservable
    .to(LifecycleObservable.converter())
```


hzsweers:
Canning this for now. If anything, will probably just adopt RxBinding's `InitialValueObservable` approach

PR #47, opened 2017-03-12
Add kotlin module

### LINKED ISSUE #7: Explore adding Kotlin modules
`autodispose-kotlin`

`autodispose-android-kotlin`

Things to consider:
- Worth it?
- What would the API look like?

hzsweers:
Resolves #7

This tacks on convenience extension functions to RxJava types, such that you can just do this:

```kotlin
myObservable
    .doWhatever()
    .autoDisposeWith(this)
    .subscribe()
```

Did some opportunistic project house cleaning as well

### DIFF
--- a/autodispose-kotlin/src/test/kotlin/com/uber/autodispose/kotlin/AutoDisposeKotlinTest.kt
+++ b/autodispose-kotlin/src/test/kotlin/com/uber/autodispose/kotlin/AutoDisposeKotlinTest.kt
@@ -60,8 +60,8 @@ class AutoDisposeKotlinTest {
     }
 
     override fun correspondingEvents(): Function<LifecycleEvent, LifecycleEvent> {
-      return Function { event ->
-        when (event) {
+      return Function {
+        when (it) {
           is Start -> End()
           is End -> throw LifecycleEndedException()
         }
@@ -74,7 +74,7 @@ class AutoDisposeKotlinTest {
         .autoDisposeWith(scopeMaybe)
         .subscribe(o)
 
-    o.assertValue { v -> v == "Hello" }
+    o.assertValue { it == "Hello" }
     o.assertComplete()
   }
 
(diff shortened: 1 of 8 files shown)


0legg:
LGTM, the only thing I have add to it — have you tried [Spek](http://spekframework.org/) for unit tests?

autodispose-kotlin/src/test/kotlin/com/uber/autodispose/kotlin/AutoDisposeKotlinTest.kt
```diff
@@ -0,0 +1,545 @@
...
+      }
+    }
+  }
+
+  @Test fun observable_maybeNormalCompletion() {
+    Observable.just("Hello")
+        .autoDisposeWith(scopeMaybe)
+        .subscribe(o)
+
+    o.assertValue { v -> v == "Hello" }
```
0legg: You can use `it` for brevity; while it's better to have named lambda parameters for more complicated logic, oneliner will look nicer with `it`

hzsweers:
screen shot 2017-03-12 at 1 45 57 am


Heh, so I actually tried spek with this, but every time I try spek something seems to go stupidly wrong.  It seems super unreliable, either with its build setup or just plain not working (exception thrown above is considered successful ¯\_(ツ)_/¯). I get a totally different error when trying to run from the command line too

autodispose-kotlin/src/test/kotlin/com/uber/autodispose/kotlin/AutoDisposeKotlinTest.kt
hzsweers: Good point - 2e047aa

Forgot this makes `when` blocks with instance checks really pretty.

```kotlin
return Function {
  when (it) {
    is Start -> End()
    is End -> throw LifecycleEndedException()
  }
}
```

0legg:
@hzsweers There are some [notes](http://spekframework.org/docs/latest/#setting-up-legacy) on enablin Spek with JUnit 4, and it looks like IDE isn't supported at this time — CLI only.

PR #54, opened 2017-03-16
Add AutoDisposeViewHolder Android component

hzsweers:
This adds a ViewHolder implementation/reference for using AutoDispose on recyclerviews. Due to the nature of how RecyclerViews work, just relying on `View` detaches is not a good enough heuristic, and instead we need to listen for view recycling changes.

The philosophy is this:

`AutoDisposeViewHolder` is the reference implementation, and it implements the `RecycleAwareViewHolder` interface. Others can use this as their base class if they want, or just copy its implementation out into theirs. As long as they implement `RecycleAwareViewHolder`, they can reuse the view recycling helpers. This is also super easy to push up into a base RecyclerView class, which is something we do.

It has two static helpers for usage:
  - `onViewRecycled`: This is a proxy for the same method on `RecyclerView.Adapter` that you can call from your adapter or a `RecyclerView.RecyclerListener` implementation.
  - `newRecyclerListener`: Convenience method that gives you a new listeners that just sends a call out to `AutoDisposeViewHolder#onViewRecycled`.

The helpers make it easy to hook up with the RecyclerView/Adapter, though ideally a future version of RecyclerView would have an additive API for adding listeners in the future that could make this simpler.

In code:

```java

myRecyclerView.setRecyclerListener(AutoDisposeViewHolder.newRecyclerListener());

// Rest of your stuff

static class ViewHolder extends AutoDisposeViewHolder {
    // Stuff
}
```

-----

The test is a bit wonky. See the doc on the `AutoDisposeViewHolderTest` class for details.

-----

For review - I recommend reviewing commit by commit. Read them in order, leave comments, etc. Each is pretty isolated and digestible, then at the end you can approve/request changes as needed (approval/request changes is _not_ per-commit).

### DIFF
--- a/autodispose-android/src/androidTest/java/com/uber/autodispose/android/AutoDisposeViewHolderTestActivity.java
+++ b/autodispose-android/src/androidTest/java/com/uber/autodispose/android/AutoDisposeViewHolderTestActivity.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright (C) 2017. Uber Technologies
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.uber.autodispose.android;
+
+import android.app.Activity;
+import android.os.Bundle;
+import android.support.v7.widget.LinearLayoutManager;
+import android.support.v7.widget.RecyclerView;
+
+public final class AutoDisposeViewHolderTestActivity extends Activity {
+  RecyclerView recyclerView;
+
+  @Override protected void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    recyclerView = new RecyclerView(this);
+    recyclerView.setId(android.R.id.primary);
+    recyclerView.setLayoutManager(new LinearLayoutManager(this));
+    setContentView(recyclerView);
+  }
+}
--- a/autodispose-android/src/androidTest/java/com/uber/autodispose/android/AutoDisposeViewHolderViewTest.java
+++ b/autodispose-android/src/androidTest/java/com/uber/autodispose/android/AutoDisposeViewHolderViewTest.java
@@ -39,6 +39,7 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 
 import static android.support.test.espresso.Espresso.onView;
+import static android.support.test.espresso.action.ViewActions.swipeDown;
 import static android.support.test.espresso.action.ViewActions.swipeUp;
 import static android.support.test.espresso.matcher.ViewMatchers.withId;
 import static com.google.common.truth.Truth.assertThat;
@@ -46,7 +47,7 @@ import static com.google.common.truth.Truth.assertThat;
 /**
  * Bit of an odd one, but this test works by flinging the list and ensuring the number of recycles
  * lines up with the number of disposals. We can't monitor recycling + immediate disposable directly
- *  because it's actually async.
+ * because it's actually async.
  */
 @RunWith(AndroidJUnit4.class) public final class AutoDisposeViewHolderViewTest {
   @Rule public final ActivityTestRule<AutoDisposeViewHolderTestActivity> activityRule =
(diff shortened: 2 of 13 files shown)


autodispose-android/src/main/java/com/uber/autodispose/android/AutoDisposeViewHolder.java
```diff
@@ -0,0 +1,74 @@
...
+
+  private MaybeSubject<Object> unbindNotifier;
+
+  public AutoDisposeViewHolder(View itemView) {
+    super(itemView);
+  }
+
+  private void onUnBind() {
+    emitUnBindIfPresent();
+  }
```
jbarr21: shouldn't you handle the re-creation of the MaybeSubject when the view holder is rebound?

autodispose-android/src/androidTest/java/com/uber/autodispose/android/AutoDisposeViewHolderTestActivity.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package com.uber.autodispose.android;
+
+import android.app.Activity;
+import android.os.Bundle;
+import android.support.v7.widget.LinearLayoutManager;
+import android.support.v7.widget.RecyclerView;
+
+public final class AutoDisposeViewHolderTestActivity extends Activity {
```
jbarr21: should this match your example activity name? ViewScopeTestActivity?

autodispose-android/src/androidTest/java/com/uber/autodispose/android/AutoDisposeViewHolderViewTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+import static android.support.test.espresso.Espresso.onView;
+import static android.support.test.espresso.action.ViewActions.swipeUp;
+import static android.support.test.espresso.matcher.ViewMatchers.withId;
+import static com.google.common.truth.Truth.assertThat;
+
+/**
+ * Bit of an odd one, but this test works by flinging the list and ensuring the number of recycles
+ * lines up with the number of disposals. We can't monitor recycling + immediate disposable directly
+ *  because it's actually async.
```
jbarr21: the recycling or disposable is async?

autodispose-android/src/androidTest/java/com/uber/autodispose/android/AutoDisposeViewHolderViewTest.java
```diff
@@ -0,0 +1,146 @@
...
+    interaction = onView(withId(android.R.id.primary));
+    viewDirtyIdler = new ViewDirtyIdlingResource(activity);
+    Espresso.registerIdlingResources(viewDirtyIdler);
+  }
+
+  @After public void tearDown() {
+    Espresso.unregisterIdlingResources(viewDirtyIdler);
+  }
+
+  @Test public void basicTest() {
```
jbarr21: can you also add a test for a fling to bottom, then fling back to top and check the count? this would ensure that behavior is correct when a viewholder is rebound

autodispose-android/src/main/java/com/uber/autodispose/android/AutoDisposeViewHolder.java
hzsweers: Can you elaborate?

autodispose-android/src/androidTest/java/com/uber/autodispose/android/AutoDisposeViewHolderViewTest.java
hzsweers: the test itself. The espresso artifact and the app itself are running on different UI threads, so it's actually not enough to call "dispose" and immediately check it.

autodispose-android/src/androidTest/java/com/uber/autodispose/android/AutoDisposeViewHolderViewTest.java
hzsweers: Good idea

hzsweers: 2ec85c9

autodispose-android/src/main/java/com/uber/autodispose/android/AutoDisposeViewHolder.java
hzsweers: Wait figured it out. Fixed in 2ec85c9

autodispose-android/src/androidTest/java/com/uber/autodispose/android/AutoDisposeViewHolderTestActivity.java
hzsweers: Oops 4e028d7

autodispose-android/src/main/java/com/uber/autodispose/android/ViewScopeProvider.java
```diff
@@ -33,7 +34,8 @@
  *      .empty();
  * </code></pre>
  */
-public class ViewScopeProvider implements LifecycleScopeProvider<ViewLifecycleEvent> {
+@UiThread
```
jbarr21: odd to see this on a class. will this just apply it to all methods in the class?

hzsweers: Yep

hzsweers:
Going to put this on hold for now, this solution for rebinds won't actually work as it would reset if you had multiple streams. Need to think more on it

hzsweers:
Abandoning this for now. At the end of the day there's no sane way to do this without depending on internal recyclerview APIs. Can maybe just put an `RxViewHolder` example up somewhere for posterity, but this isn't possible in a safe public API right now

Issue #52, opened 2017-03-15
LifecycleEndException handling


hzsweers:
Specifically, some might want to just dispose quietly (in production, or always) rather than error. Should we make this more configurable? If so, how/to what degree? Moving this discussion here from an internal one.

Currently - `LifecycleEndException` is propagated to the delegate observer's `onError`. I see two options:

A. If the delegate doesn't implement `onError`, it goes to `RxJavaPlugins` and one could filter that in a plugin hook. This would not cover cases where `onError` is implemented though. Is this fine?
B. We make this more explicitly configurable
  - Scoper overloads?
  - `RxJavaPlugins`-esque plugin system?
  - Nothing, advise just writing a custom `ScopeProvider` implementation using `ScopeUtils`?

CC @tonycosentini @AttwellBrian

tonycosentini:
What about just advising people to potentially non-fatal them in their own `onError` plugin implementation?

AttwellBrian:
> A. If the delegate doesn't implement onError, it goes to RxJavaPlugins and one could filter that in a plugin hook. This would not cover cases where onError is implemented though. Is this fine?

I can't think of a reason this wouldn't be fine. It was good enough for us? Do other companies implement `onError()` more than us?

tonycosentini:
After syncing with @hzsweers in person, one idea is a plugin system for AutoDispose itself. The default implementation would just route to onError like it does currently, but would allow for custom handlers (for logging/non-fatals or whatever else makes sense for a specific app).

hzsweers:
Cool, plugin system sounds good. @tonycosentini do you want to start a PR around that? Structure would probably be best in the same vein as `RxJavaPlugins`

tonycosentini:
Yup, will give it a shot.

hzsweers:
Done in #57

PR #57, opened 2017-03-21
Add plugin API for handling OutsideLifecycleExceptions.

tonycosentini:
(Addresses https://github.com/uber/AutoDispose/issues/52)

The API is very similar to RxJavaPlugins.

One thing I'm a bit concerned about here - if you're providing your own `ScopeProvider` or `LifecycleScopeProvider`, you need to either manually use `ScopeUtil` or manually call the plugin yourself before throwing anything. It would be nicer if AutoDispose does all of this for you.

### DIFF
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDisposePlugins.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDisposePlugins.java
@@ -24,10 +24,10 @@ import io.reactivex.plugins.RxJavaPlugins;
 /**
  * Utility class to inject handlers to certain standard AutoDispose operations.
  */
-public class AutoDisposePlugins {
+public final class AutoDisposePlugins {
 
     @Nullable
-    static volatile Consumer<? super OutsideLifecycleException> outsideLifecycleHandler;
+    private static volatile Consumer<? super OutsideLifecycleException> outsideLifecycleHandler;
 
     /**
      * Prevents changing the plugins.
@@ -36,7 +36,8 @@ public class AutoDisposePlugins {
 
     /**
      * Prevents changing the plugins from then on.
-     * <p>This allows container-like environments to prevent client messing with plugins.</p>
+     *
+     * This allows container-like environments to prevent client messing with plugins.
      */
     public static void lockdown() {
         lockdown = true;
@@ -67,30 +68,6 @@ public class AutoDisposePlugins {
             throw new IllegalStateException("Plugins can't be changed anymore");
         }
         outsideLifecycleHandler = handler;
-    }
-
-    /**
-     * Called when an outside lifecycle error occurs.
-     * @param error the error to report
-     */
-    public static void onOutsideLifecycleException(@NonNull OutsideLifecycleException error) {
-        Consumer<? super  OutsideLifecycleException> f = outsideLifecycleHandler;
-
-        if (error == null) {
-            RxJavaPlugins.onError(
-                    new NullPointerException("onOutsideLifecycleException called with null. Null values are generally "
-                    + "not allowed in 2.x operators and sources.")
-            );
-        }
-
-        if (f != null) {
-            try {
-                f.accept(error);
-                return;
-            } catch (Throwable e) {
-                RxJavaPlugins.onError(e);
-            }
-        }
     }
 
     /**
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDisposePlugins.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDisposePlugins.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright (c) 2017. Uber Technologies
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.uber.autodispose;
+
+import io.reactivex.annotations.NonNull;
+import io.reactivex.annotations.Nullable;
+import io.reactivex.functions.Consumer;
+import io.reactivex.plugins.RxJavaPlugins;
+
+/**
+ * Utility class to inject handlers to certain standard AutoDispose operations.
+ */
+public class AutoDisposePlugins {
+
+    @Nullable
+    static volatile Consumer<? super OutsideLifecycleException> outsideLifecycleHandler;
+
+    /**
+     * Prevents changing the plugins.
+     */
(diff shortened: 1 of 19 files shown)


autodispose/src/main/java/com/uber/autodispose/AutoDisposePlugins.java
```diff
@@ -0,0 +1,102 @@
...
+
+import io.reactivex.annotations.NonNull;
+import io.reactivex.annotations.Nullable;
+import io.reactivex.functions.Consumer;
+import io.reactivex.plugins.RxJavaPlugins;
+
+/**
+ * Utility class to inject handlers to certain standard AutoDispose operations.
+ */
+public class AutoDisposePlugins {
```
hzsweers: `final`

autodispose/src/main/java/com/uber/autodispose/AutoDisposePlugins.java
```diff
@@ -0,0 +1,102 @@
...
+    static volatile Consumer<? super OutsideLifecycleException> outsideLifecycleHandler;
+
+    /**
+     * Prevents changing the plugins.
+     */
+    static volatile boolean lockdown;
+
+    /**
+     * Prevents changing the plugins from then on.
+     * <p>This allows container-like environments to prevent client messing with plugins.</p>
```
hzsweers: nit - nextline the text, no need for the closing paragraph marker

autodispose/src/main/java/com/uber/autodispose/AutoDisposePlugins.java
```diff
@@ -0,0 +1,102 @@
...
+        Consumer<? super  OutsideLifecycleException> f = outsideLifecycleHandler;
+
+        if (error == null) {
+            RxJavaPlugins.onError(
+                    new NullPointerException("onOutsideLifecycleException called with null. Null values are generally "
+                    + "not allowed in 2.x operators and sources.")
+            );
+        }
+
+        if (f != null) {
```
hzsweers: What if `f` is null?

autodispose/src/main/java/com/uber/autodispose/ScopeUtil.java
```diff
@@ -74,15 +75,25 @@ private ScopeUtil() {
       @Override public MaybeSource<? extends LifecycleEndNotification> call() throws Exception {
         E lastEvent = provider.peekLifecycle();
         if (checkStartBoundary && lastEvent == null) {
-          throw new LifecycleNotStartedException();
+          LifecycleNotStartedException exception = new LifecycleNotStartedException();
+          if (AutoDisposePlugins.outsideLifecycleHandler != null) {
+            AutoDisposePlugins.outsideLifecycleHandler.accept(exception);
```
hzsweers: Shouldn't this use the API rather than the fields directly?

autodispose/src/main/java/com/uber/autodispose/ScopeUtil.java
```diff
@@ -74,15 +75,25 @@ private ScopeUtil() {
...
         }
         E endEvent;
         try {
           endEvent = provider.correspondingEvents()
               .apply(lastEvent);
         } catch (Exception e) {
           if (checkEndBoundary && e instanceof LifecycleEndedException) {
-            throw e;
+            if (AutoDisposePlugins.outsideLifecycleHandler != null) {
+              AutoDisposePlugins.outsideLifecycleHandler.accept((LifecycleEndedException) e);
```
hzsweers: Same here with using the API

autodispose/src/main/java/com/uber/autodispose/ScopeUtil.java
```diff
@@ -74,15 +75,25 @@ private ScopeUtil() {
...
+              return Maybe.empty();
```
hzsweers: Shouldn't this emit immediately to dispose immediately?

autodispose/src/main/java/com/uber/autodispose/ScopeUtil.java
hzsweers: Also shouldn't we do something based if we proceed past accept? Like dispose or something?

hzsweers:
I think `ScopeUtil` is an ok place for this. It's there to be a utility for people to use, and we'll support it in the first party implementation. Kind of like RxJava plugins - people can reuse them but only built-ins are guaranteed to use them

autodispose/src/main/java/com/uber/autodispose/ScopeUtil.java
hzsweers: Blocking on this one

hzsweers:
Wait I'm dumb. `Maybe.empty()` is fine, I was thinking of `Maybe.never()`. It does need to emit something though, we actually don't do anything on empty currently (but maybe we should?)

hzsweers:
Ping on this

autodispose/src/main/java/com/uber/autodispose/AutoDisposePlugins.java
tonycosentini: Turns out, this method isn't even used, I'm going to remove it.

autodispose/src/main/java/com/uber/autodispose/AutoDisposePlugins.java
```diff
@@ -36,7 +36,8 @@
 
     /**
      * Prevents changing the plugins from then on.
-     * <p>This allows container-like environments to prevent client messing with plugins.</p>
+     *
```
hzsweers: Keep the `<p>`

autodispose/src/test/java/com/uber/autodispose/AutoDisposeSubscriberTest.java
```diff
@@ -179,6 +185,25 @@
...
+      }
+    });
+    BehaviorSubject<Integer> lifecycle = BehaviorSubject.create();
+    TestSubscriber<Integer> o = new TestSubscriber<>();
+    LifecycleScopeProvider<Integer> provider = TestUtil.makeLifecycleProvider(lifecycle);
+    Flowable.just(1)
+            .to(new FlowableScoper<Integer>(provider))
+            .subscribe(o);
+
+    o.assertNotSubscribed();
```
tonycosentini: @hzsweers any idea why this would fail? This holds true for all the other types but for Flowable, it doesn't seem like it's the case.

hzsweers: TestSubscriber and TestObserver are special/weird cases where they never unsubscribe. I'd just save the Disposable off and verify it's disposed.

hzsweers:
I think the tests are structured in an odd way. At the end of the day, the behavior is what we're interested in. The way that gets done might be an implementation detail, which this kind of verifies manually now (kind of like mockito but not quite). This is also not a conventional use case of `RecordingObserver`, since a handler is not really an `Observer`. 

What would you think of making it a more functional test? E.g. 

- set a handler that silently swallows `LifecycleEndedException` exceptions and verify your stream is just disposed without errors
- a handler that swallows `LifecycleNotStartedException` exceptions and verify the stream is disposed without errors
- a handler that rethrows an alternate exception and verify the stream errored with that exception

Let me know what you think

hzsweers:
Also can you add a bit to the README?

PR #61, opened 2017-04-29
Proposal - reintroduce static scoper factories

hzsweers:
These don't autocomplete nicely, but they're sort of nicer to read in code. Want to put this up for discussion to see if there's interest in having them as an option (albeit `@Experimental`).

Now you'd do

```java
myObservable
    .to(AutoDipose.<Stuff>observableScoper(this)
    .subscribe()
```

### DIFF
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDispose.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDispose.java
@@ -1,0 +1,74 @@
+package com.uber.autodispose;
+
+import io.reactivex.Maybe;
+import io.reactivex.annotations.Experimental;
+
+/**
+ * Static factories for creating new scopers.
+ */
+@Experimental public final class AutoDispose {
+
+  private AutoDispose() {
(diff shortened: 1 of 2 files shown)


tonycosentini:
I'm pretty indifferent to this.

Benefits to this seem to be:
* Less objects created (although probably not enough here to make it worth it?)

Downsides seem to be:
* More characters
* (Subjective) having an instance based + static API side by side is a bit confusing

hzsweers:
side by side for sure is a bit confusing. 

I don't follow the more characters bit though?

tonycosentini:
It's a bit subjective, but some people tend to not import static methods, so class name + static method name is a bit longer than `new ScoperClass(...)`.

That's not a huge concern - but I do think having two APIs is pretty confusing. In addition since there is already some traction here, it doesn't seem like a breaking change with no clear benefit would make a ton of sense in my opinion.

hzsweers:
Closing in favor of #75. If this were implemented, it'd be in that

PR #65, opened 2017-05-03
Lower support annotations

hzsweers:
This is the earliest official version I could find in the changelog. It's useful for a library to only maintain the minimum supported support library/sdk version since otherwise consumers have to do a bunch of suppression hackery.


hzsweers:
Actually we use RestrictTo

PR #69, opened 2017-05-12
Switch to provided jsr305 and nullability improvements

hzsweers:
This takes inspiration from Square's recent blog post and brings back the jsr305 dependency as a provided dependency. Benefits are static analysis and kotlin code still can leverage this.

Main functional differences are annotating a `package-info` with `@ParametersAreNonnullByDefault` annotation and switching to javax' `@Nullable` annotation instead of RxJava's.

Thank you for contributing to AutoDispose. Before pressing the "Create Pull Request" button, please consider the following points: 

Prior reading:
  - https://github.com/square/okio/pull/307
  - https://medium.com/square-corner-blog/rolling-out-nullable-42dd823fbd89

PR #70, opened 2017-05-14
More nullability annotation work

hzsweers:
This is a followup to #69, and improves it in two main ways:
- This adds it to subpackages as well, which I missed in the first pass.
- Switches to a custom `PackageNonNull` annotation instead, which expands the checks to also include return types, fields, local vars, etc. It's pretty broad right now but we could reduce scope if need be. This is pulled into a separate artifact for re-use, but still only applied as compileOnly. `@ParametersAreNonnullByDefault` actually only covers method parameters it seems.

### DIFF
--- a/autodispose-provided/gradle.properties
+++ b/autodispose-provided/gradle.properties
@@ -1,0 +1,3 @@
+POM_NAME=AutoDispose (Provided)
+POM_ARTIFACT_ID=autodispose-provided
+POM_PACKAGING=jar
--- a/autodispose-provided/gradle.properties
+++ b/autodispose-provided/gradle.properties
@@ -1,3 +1,0 @@
-POM_NAME=AutoDispose (Provided)
-POM_ARTIFACT_ID=autodispose-provided
-POM_PACKAGING=jar
--- a/autodispose-provided/src/main/java/com/uber/autodispose/internal/EverythingNonNullByDefault.java
+++ b/autodispose-provided/src/main/java/com/uber/autodispose/internal/EverythingNonNullByDefault.java
@@ -21,5 +21,5 @@ import javax.annotation.meta.TypeQualifierDefault;
     ElementType.METHOD,
     ElementType.PARAMETER
 })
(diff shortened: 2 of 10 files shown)


autodispose-provided/gradle.properties
```diff
@@ -0,0 +1,3 @@
+POM_NAME=AutoDispose (Provided)
```
kageiit: why are we shipping this to maven? this is compileOnly

hzsweers: Still needs to be available in sources jar for static analysis right?

msridhar:
LGTM.  But unless you have public fields, the main change here is that the annotation means return types are `@NonNull` by default, right?

autodispose-provided/src/main/java/com/uber/autodispose/internal/EverythingNonNullByDefault.java
```diff
@@ -0,0 +1,25 @@
...
+ * - There is a default parameter annotation applied to a more tightly nested element.
+ */
+@Documented
+@Nonnull
+@TypeQualifierDefault({
+    ElementType.FIELD,
+    ElementType.METHOD,
+    ElementType.PARAMETER
+})
+@Retention(RetentionPolicy.RUNTIME)
```
kageiit: SOURCE retention please. Since this is not really shipped out to consumers

hzsweers:
the annotation applies to parameters, method return types, and fields in general, even private ones (though those only really matter for internal details)

autodispose-provided/src/main/java/com/uber/autodispose/internal/EverythingNonNullByDefault.java
hzsweers: I coped this from `@ParametersAreNonnullByDefault`, let me see if I can find the justification

hzsweers: 3dcc0a0

Issue #37, opened 2017-02-08
Add demo scope provider for Fragment


tonycosentini:

tadfisher:
`FragmentLifecycleCallbacks` exists for the support library.

hzsweers:
Same issues as https://github.com/uber/AutoDispose/issues/36#issuecomment-288888092 I think, but I'm not too familiar with the fragment callbacks. Is there a way to query the current state?

hzsweers:
Closing in favor of #71

PR #71, opened 2017-05-24
Arch components lifecycle support

### LINKED ISSUE #36: Add demo scope provider for Activity

hzsweers:
This is a first pass of support for Android Architecture Components' `Lifecycle` support. This works fairly similarly to `ViewScopeProvider` from a structural standpoint. `LifecycleOwner` or `Lifecycle` can be passed in, an inner observer is passed in and removed upon disposal.

Thoughts for discussion:
  - Name is up for grabs. This one's a bit long :)
  - Should it be its own artifact? I'm leaning toward this, then we could also do an alpha release of it separately.
  - Things get a bit convoluted for me around `State` vs. `Event`. I'm not sure which one we should watch for, but all the querying APIs revolve around `State` so I went with that.
  - Does it matter if it's on the main thread?

If this looks good - I'll take a pass at tests.

I think this would also resolve #36 and #37 by just supporting them (in the support lib anyway)


autodispose-android/src/main/java/com/uber/autodispose/android/LifecycleStatesObservable.java
```diff
@@ -0,0 +1,72 @@
...
+      this.observer = observer;
+    }
+
+    @Override protected void onDispose() {
+      lifecycle.removeObserver(this);
+    }
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_ANY) void onStateChange() {
+      if (!isDisposed()) {
+        observer.onNext(lifecycle.getCurrentState());
```
hzsweers: I'm not sure if this could result in duplicate state emissions since this fires on events. Will find out

autodispose-android/build.gradle
```diff
@@ -49,6 +49,7 @@ dependencies {
   compile deps.rx.java
   compile deps.rx.android
   compile deps.support.annotations
+  compile "android.arch.lifecycle:extensions:1.0.0-alpha1"
```
tonycosentini: +1 for a separate module approach

autodispose-android/src/main/java/com/uber/autodispose/android/LifecycleOwnerScopeProvider.java
```diff
@@ -0,0 +1,98 @@
...
+ */
+public final class LifecycleOwnerScopeProvider implements LifecycleScopeProvider<Lifecycle.State> {
+
+  private static final Function<Lifecycle.State, Lifecycle.State> CORRESPONDING_EVENTS =
+      new Function<Lifecycle.State, Lifecycle.State>() {
+        @Override public Lifecycle.State apply(Lifecycle.State lastEvent) throws Exception {
+          switch (lastEvent) {
+            case INITIALIZED:
+              throw new LifecycleNotStartedException();
+            case CREATED:
```
tonycosentini: Is the lifecycle intentionally not 1-1? It seems very strange to have multiple start events map to the same stop event.

hzsweers: There's only one stop event, not sure why

autodispose-android/build.gradle
yigit: +1. at least until we call it beta (which means stable api)

autodispose-android/src/main/java/com/uber/autodispose/android/LifecycleOwnerScopeProvider.java
yigit: I think you are confusing states vs events. 
https://developer.android.com/topic/libraries/architecture/lifecycle.html

e.g. both onStart and onPause brings you to the STARTED state. I'm not sure where you are using these but this seems wrong.

autodispose-android/build.gradle
hzsweers: yup moved it

jaychang0917:
I think `Event` is more suitable since you can't manage to cancel subscription in corresponding lifecycle event (e.g. `onResume` -> `onPause`) if you use `State`.

hzsweers:
Yep event is the right way to go. Sorry for the lag on this, kind of mentally put it on the backburner since it was still in alpha. Will jump on it this coming week

hzsweers:
So I'm a little stuck on this right now. I've switched the implementation to events, but I'm not sure how to test this. I tried writing this as an espresso test (similar to how the `-android` artifact is), but lifecycle events when manipulating activities in espresso seem to fire events in random order.

yigit:
fyi, you don't need to keep this as a separate artifact, Lifecycle's core libs will be a dependency of support Fragments so all apps will have them.

autodispose-android/build.gradle
yigit: btw, you don't need extensions, you should only need lifecycle:common

hzsweers:
yeah I was just making it separate for while it was in alpha, but maybe you'll beat me to it :)

autodispose-android/build.gradle
hzsweers: cool, I'll remove it

hzsweers: c8ad885

hzsweers:
Will handle rebasing before landing

autodispose-android/build.gradle
hzsweers: Actually I remember now why I added it, I need `LifecycleRegistry` for the `TestAndroidLifecycleScopeProvider` implementation. Added back in 1c06bef

hzsweers:
Ok, things should be ready to go here. Going to do a rebase

hzsweers:
Rebased. Requesting review from @tonycosentini and @jbarr21, also welcome feedback from anyone else on the thread :)

autodispose-android-archcomponents/src/androidTest/java/com/uber/autodispose/android/archcomponents/AndroidLifecycleScopeProviderTest.java
```diff
@@ -0,0 +1,143 @@
...
+    assertThat(o.takeNext()).isEqualTo(1);
+
+    lifecycle.emit(Lifecycle.Event.ON_PAUSE);
+    lifecycle.emit(Lifecycle.Event.ON_STOP);
+    lifecycle.emit(Lifecycle.Event.ON_DESTROY);
+
+    subject.onNext(2);
+    o.assertNoMoreEvents();
+
+    d.dispose();
```
tonycosentini: Dumb question - why is this needed?

autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/archcomponents/LifecycleEventsObservable.java
```diff
@@ -0,0 +1,105 @@
...
+import static com.uber.autodispose.android.internal.AutoDisposeAndroidUtil.isMainThread;
+
+@RestrictTo(LIBRARY) class LifecycleEventsObservable extends Observable<Event> {
+
+  private final Lifecycle lifecycle;
+  private final BehaviorSubject<Event> eventsObservable = BehaviorSubject.create();
+
+  @SuppressWarnings("CheckReturnValue") LifecycleEventsObservable(Lifecycle lifecycle) {
+    this.lifecycle = lifecycle;
+    // Backfill if already created for boundary checking
```
tonycosentini: nit: doesn't seem like it matters with today's implementation - but would it ever be beneficial to replay all the previous events?

autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/archcomponents/LifecycleEventsObservable.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  Event getValue() {
+    return eventsObservable.getValue();
+  }
+
+  @Override protected void subscribeActual(Observer<? super Event> observer) {
+    if (!isMainThread()) {
+      observer.onError(
+          new IllegalStateException("Lifecycles can only be bound to on the main thread!"));
```
tonycosentini: Is this a strict requirement with the new architecture library?

hzsweers: Not necessarily but it's always emitted on the main thread and you could have a race condition if you don't.

autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/archcomponents/LifecycleEventsObservable.java
hzsweers: I'm not sure what you mean. Like having that be a use case of the library?

autodispose-android-archcomponents/src/androidTest/java/com/uber/autodispose/android/archcomponents/AndroidLifecycleScopeProviderTest.java
hzsweers: I don’t remember (this test is largely a copy of the viewscopeprovidertest), I think it’s not and I just did it out of habit or something. Will try removing

hzsweers: d5eff71

autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/archcomponents/LifecycleEventsObservable.java
hzsweers: Checked with @yigit, answer is yes otherwise no guarantees

hzsweers:
Rebased and ran UI tests locally

PR #79, opened 2017-09-07
Remove provided module in favor of javax-extras

hzsweers:
This removes the internal `provided` module in favor of the now-public `javax-extras` artifact with its helper parameters.


BanzaiMan:
@hzsweers invited us to comment (https://github.com/travis-ci/travis-ci/issues/8360#issuecomment-327865269)

I am not quite sure what is happening here. In https://travis-ci.org/uber/AutoDispose/builds/272986414#L1139-L1144, you can see that `build-tools 26.0.1` is installed (by `android update sdk`, which may or may not make a difference). According to https://developer.android.com/studio/command-line/avdmanager.html, one should expect this in `android_sdk/tools/bin/`, but it is not clear to me where this is. Our default installation exists in `/usr/local/android-sdk` (which is a symlink to `/usr/local/android-sdk-25.2.3`), but after the installation `avdmanager` does not exist in `/usr/local/android-sdk/tools/bin`.

hzsweers:
@BanzaiMan this seems like a pretty major issue if no UI tests running on travis work now right?

BanzaiMan:
@hzsweers Yes, I agree that this is a major problem. It really is not clear to me why the installed components seem to disappear.

PR #89, opened 2017-09-21
Add support for exposing delegate observers.

### LINKED ISSUE #78: Expose the underlying observer in AutoDisposeObservers
So consumers could get access to changes like this https://github.com/ReactiveX/RxJava/pull/5590

mswysocki:
Resolves #78

This change updated `AutoDisposingSubscriber` implement `FlowableSubscriber` instead of `Subscriber`. We wanted to have more consistency with how the rest of the `Observer`s are handled in RxJava. This change remains backward compatible because `FlowableSubscriber` still extends `Subscriber`.

This also exposes all the different implementations of `Observer` to allow for introspection to the delegate `Observer`. This will be particularly useful for https://github.com/ReactiveX/RxJava/pull/5590.

### DIFF
--- a/autodispose/src/main/java/com/uber/autodispose/observers/AutoDisposingObserver.java
+++ b/autodispose/src/main/java/com/uber/autodispose/observers/AutoDisposingObserver.java
@@ -26,7 +26,7 @@ import io.reactivex.disposables.Disposable;
 public interface AutoDisposingObserver<T> extends Observer<T>, Disposable {
 
     /**
-     * @return {@link Observer} The delegate Observer that is used under the hood for introspection purposes.
+     * @return The delegate {@link Observer} that is used under the hood for introspection purposes.
      */
     Observer<? super T> delegateObserver();
 }
--- a/autodispose/src/main/java/com/uber/autodispose/observers/AutoDisposingObserver.java
+++ b/autodispose/src/main/java/com/uber/autodispose/observers/AutoDisposingObserver.java
@@ -17,6 +17,7 @@
 package com.uber.autodispose.observers;
 
 import io.reactivex.Observer;
+import io.reactivex.annotations.Experimental;
 import io.reactivex.disposables.Disposable;
 
 /**
@@ -26,7 +27,9 @@ import io.reactivex.disposables.Disposable;
 public interface AutoDisposingObserver<T> extends Observer<T>, Disposable {
 
   /**
-   * @return The delegate {@link Observer} that is used under the hood for introspection purposes.
+   * @return The delegate {@link Observer} that is used under the hood for introspection purpose.
+   * This will be updated once LambdaIntrospection is out of @Experimental in RxJava.
    */
+  @Experimental
   Observer<? super T> delegateObserver();
 }
--- a/autodispose/src/main/java/com/uber/autodispose/observers/AutoDisposingObserver.java
+++ b/autodispose/src/main/java/com/uber/autodispose/observers/AutoDisposingObserver.java
@@ -25,8 +25,8 @@ import io.reactivex.disposables.Disposable;
  */
 public interface AutoDisposingObserver<T> extends Observer<T>, Disposable {
 
-    /**
-     * @return The delegate {@link Observer} that is used under the hood for introspection purposes.
-     */
-    Observer<? super T> delegateObserver();
+  /**
+   * @return The delegate {@link Observer} that is used under the hood for introspection purposes.
+   */
+  Observer<? super T> delegateObserver();
 }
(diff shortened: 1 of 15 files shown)


autodispose/src/main/java/com/uber/autodispose/observers/AutoDisposingSubscriber.java
```diff
@@ -24,4 +25,4 @@
  * A {@link Disposable} {@link Subscriber} that can automatically dispose itself. Interface here
  * for type safety but enforcement is left to the implementation.
  */
-public interface AutoDisposingSubscriber<T> extends Subscriber<T>, Subscription, Disposable {}
+public interface AutoDisposingSubscriber<T> extends FlowableSubscriber<T>, Subscription, Disposable {}
```
hzsweers: Nit: would be good to include a tidbit of the explanation you put in the PR description in this commit message

autodispose/src/main/java/com/uber/autodispose/observers/AutoDisposingObserver.java
```diff
@@ -23,4 +23,10 @@
  * A {@link Disposable} {@link Observer} that can automatically dispose itself.
  * Interface here for type safety but enforcement is left to the implementation.
  */
-public interface AutoDisposingObserver<T> extends Observer<T>, Disposable {}
+public interface AutoDisposingObserver<T> extends Observer<T>, Disposable {
+
+    /**
+     * @return {@link Observer} The delegate Observer that is used under the hood for introspection purposes.
```
hzsweers: Nit: Did you mean to link the `Observer` here rather than inline?

I'd expect this to be like `@return The delegate {@link Observer} that is used under the hood for introspection purposes.`

autodispose/src/test/java/com/uber/autodispose/AutoDisposeObserverTest.java
```diff
@@ -256,6 +264,31 @@
...
+          return observer;
+        }
+      });
+      Observable.just(1).to(new ObservableScoper<Integer>(Maybe.never())).subscribe();
+
+      assertThat(atomicAutoDisposingObserver.get()).isNotNull();
+      assertThat(atomicAutoDisposingObserver.get()).isInstanceOf(AutoDisposingObserver.class);
+      assertThat(((AutoDisposingObserver)atomicAutoDisposingObserver.get()).delegateObserver()).isNotNull();
+      assertThat(((AutoDisposingObserver)atomicAutoDisposingObserver.get()).delegateObserver())
+              .isEqualTo(atomicObserver.get());
```
hzsweers: This should be an instance check, not equality

autodispose/src/test/java/com/uber/autodispose/AutoDisposeObserverTest.java
```diff
@@ -17,26 +17,34 @@
...
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import org.junit.After;
 import org.junit.Test;
 
 import static com.google.common.truth.Truth.assertThat;
 
 public class AutoDisposeObserverTest {
 
+  private final AtomicReference<Observer> atomicObserver = new AtomicReference();
```
hzsweers: since these are only used in one test, let's make them `final` local variables in those methods instead

autodispose/src/test/java/com/uber/autodispose/AutoDisposeObserverTest.java
```diff
@@ -154,7 +162,7 @@
     lifecycle.onNext(1);
     source.onNext(2);
 
-    assertThat(source.hasObservers()).isTrue();
+    assertThat(source.hasObservers()).isTrue();Å
```
hzsweers: wat.gif

mswysocki: I believe this was from you trying to use my keyboard.

autodispose/src/test/java/com/uber/autodispose/AutoDisposeObserverTest.java
mswysocki: Does it even make sense to have these as member variables? They probably could just be defined in the test that they are being used.

mswysocki:
- Let's inline the {@link Observer} mentions in docs, it should read like a sentence since they are linkified inline. Make sure you link the correct observer as well (e.g. right now they all use Observer but there is a unique Observer for each type, like SingleObserver). e3dd21a
- Let's add a comment to the docs that explicitly call out that this API will change once the LambdaIntrospection stuff is out of @Experimental in RxJava (at which point we'll update it to return that interface instead and mark it nullable to return null upon non conforming types). bda1040
- Let's add @Experimental to all the methods to indicate their plans to change. Borrow the RxJava annotation. bda1040
- Let's make the atomic references in tests local to their specific unit tests rather than fields since they're only used in those methods. 7270758
- Use reference comparison for the delegate observer in tests, not isEqualTo (which checks equality and isn't what we want). 45e6c31

PR #94, opened 2017-09-28
Rename ArchComponents to AndroidLifecycle.

### LINKED ISSUE #91: Rename ArchComponents stuff to AndroidLifecycle
A little simpler. We can add an overload for views too out of convenience.

lsvijay:
Rename ArchComponents to AndroidLifecycle.
Summary:
* Rename com.uber.autodispose.android.archcomponents package to com.uber.autodispose.android.lifecycle
* Rename AndroidLifecycleScopeProvider to just AndroidLifecycle

Resolves #91

### DIFF
--- a/autodispose-android-archcomponents/src/androidTest/java/com/uber/autodispose/android/androidlifecycle/AndroidLifecycleScopeProviderTest.java
+++ b/autodispose-android-archcomponents/src/androidTest/java/com/uber/autodispose/android/androidlifecycle/AndroidLifecycleScopeProviderTest.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.uber.autodispose.android.archcomponents;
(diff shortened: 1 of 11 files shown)


hzsweers:
Thanks!

hzsweers:
Soooo after looking at this more and seeing it next to other usages, I think we should actually keep the `ScopeProvider`. It is a bit wordy, but it's become a pattern now. It does make it clear what it's for too

lsvijay:
Agree, sounds good to me.

On Oct 1, 2017, at 12:50 AM, Zac Sweers <notifications@github.com<mailto:notifications@github.com>> wrote:


Soooo after looking at this more and seeing it next to other usages, I think we should actually keep the ScopeProvider. It is a bit wordy, but it's become a pattern now. It does make it clear what it's for too

PR #96, opened 2017-09-29
Various arch components support fixes

### LINKED ISSUE #93: AndroidLifecycleScopeProvider do not work
I tried the `AndroidLifecycleScopeProvider` implementation with the following code, but it doesn't work and keep emitting when onStop()` is called.

```kotlin
override fun onStart() {
    super.onStart()
    Observable.interval(0, 1, TimeUnit.SECONDS)
      .doOnDispose { println("dispose") }
      .to(ObservableScoper(AndroidLifecycleScopeProvider.from(this)))
      .subscribe {
        println("emit:" + it)
      }
  }

  override fun onStop() {
    super.onStop()
    println("onStop")
  }
```

Thanks.

hzsweers:
This is a few miscellaneous arch components fixes:
- Enabled releases now
- Fix a deprecated implementation to use `LifecycleOwner`
- Make tests more comprehensive and realistic
- Fix #93 by offsetting backfilling up one. See issue there for more info, but the gist is that lifecycle `State`s are always the preceding event until after the callback is completed. A little confusing, but we can hide the weirdness under the hood as implementation detail.

### DIFF
--- a/autodispose-android-archcomponents/src/androidTest/java/com/uber/autodispose/android/lifecycle/AndroidLifecycleTest.java
+++ b/autodispose-android-archcomponents/src/androidTest/java/com/uber/autodispose/android/lifecycle/AndroidLifecycleTest.java
@@ -42,18 +42,17 @@ import static com.google.common.truth.Truth.assertThat;
 
   @Rule public UiThreadTestRule uiThreadTestRule = new UiThreadTestRule();
 
-  @Test @UiThreadTest public void observable_normal() {
+  @Test @UiThreadTest public void observable_beforeCreate() {
     final RecordingObserver<Integer> o = new RecordingObserver<>(LOGGER);
     final PublishSubject<Integer> subject = PublishSubject.create();
 
     // Spin it up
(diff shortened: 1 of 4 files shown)


autodispose-android-archcomponents/src/androidTest/java/com/uber/autodispose/android/lifecycle/AndroidLifecycleTest.java
```diff
@@ -75,38 +71,112 @@
     o.assertNoMoreEvents();
   }
 
-  @Test public void observable_offMainThread_shouldFail() {
-    RecordingObserver<Integer> o = new RecordingObserver<>(LOGGER);
-    PublishSubject<Integer> subject = PublishSubject.create();
+  @Test @UiThreadTest public void observable_createDestroy() {
```
AttwellBrian: Why `UiThreadTest`?

hzsweers:
> Why crash when off background thread?

[Offensive programming](https://en.wikipedia.org/wiki/Offensive_programming), arch components lifecycle handling only supports main thread (including `@MainThread` annotations on the methods).

Issue #102, opened 2017-10-08
Removing synthetic accessors


bangarharshit:
There is around 27 synthetic accessors method (most of which are in tests).  RecordingObserver, AutoDisposingSubscriberImpl (and similar classes) has around 12 methods. Maybe convert them to package protected to reduce dex count by 12.

hzsweers:
want to send a proposal PR? Can't promise all of them for API reasons but would want to see what the scope is.

bangarharshit:
Created a PR. Also verified that proguard is not inling the methods.

bangarharshit:
Closing since it is merged. All the other synthetic accessors are in tests and which can be safely ignored.

PR #103, opened 2017-10-08
Removed synthentic accessor from auto disposables and recording observer

bangarharshit:
Removed synthetic accessors.  Fixes - https://github.com/uber/AutoDispose/issues/102
Total dex count 
- autodispose package: Master - 454, New - 449, 
- recording observer - Master - 49, New - 42

### DIFF
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java
@@ -81,7 +81,8 @@ final class AutoDisposingCompletableObserverImpl implements AutoDisposingComplet
     }
   }
 
-  private void callMainSubscribeIfNecessary(Disposable d) {
+  /* private */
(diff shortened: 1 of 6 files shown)


hzsweers:
This seems fine to me. Mind adding a comment on those methods to indicate that they're normally `private` but made package-local for the accessors?

Just adding something like an inline `/* private */` might be enough context for now. I've proposed an annotation in error-prone for enforcement of this - https://github.com/google/error-prone/issues/765

bangarharshit:
Added comments to indicate the scope.

bangarharshit:
Sorry. Making it inline.

hzsweers:
Make sure you sign the correct CLA, it seems you're committing from a different github account

bangarharshit:
Fixed the account for commit.

bangarharshit:
Sure. Will take care of it. 

IntelliJ provides inspection for these but yeah some kind of error-prone check on CI will be great. Maybe a custom checker if the repo supports that.

PR #111, opened 2017-10-16
Structured Android components (and some refactorings)

### LINKED ISSUE #87: Proposal: More structured android components
Currently have -android and -android-archcomponents

Wondering if we should move these under a `/android` directory and do the following

- Provide kotlin components of existing ones with extension functions (`View.scope()` rather than `ViewScopeProvider.from(view)`)
- Provide ready-to-use android components like `AutoDisposeActivity` or `AutoDisposeViewHolder`
- Possibly separate out the `TestAndroidLifecycle` to its own artifact so the main archcomponents artifact only have a runtime dependency

### LINKED ISSUE #112: Consider bundling a proguard configuration file with Android components
AutoDipose 0.3.0 requires something like this in my project (fwiw, 0.2.0 didn't need anything):
```
-dontwarn com.uber.javaxextras.FieldsMethodsAndParametersAreNonNullByDefault
```

### LINKED ISSUE #113: Compatibility with android.arch.lifecycle 1.0.0-rc1
Looks like AutoDispose 0.3.0 is compatible only with android.arch.lifecycle 1.0.0-beta1 and the move to rc1 is not backwards compatible:
```
java.lang.ClassCastException: com.uber.autodispose.android.lifecycle.LifecycleEventsObservable_ArchLifecycleObserver_LifecycleAdapter cannot be cast to android.arch.lifecycle.GeneratedAdapter
```

This is probably fixed by the bump to use rc1 in PR #111. 

This issue can serve as a sentry until that PR Is committed and a new version is released.

hzsweers:
This is a somewhat big PR, but isolated commits

Overview:
- Move android components to `android/`
- Couple opportunistic fixes
- Move `TestAndroidLifecycleScopeProvider` to its own `autodispose-android-archcomponents-test` artifact, rename to more appropriate `TestLifecycleOwner`.
  - This is useful because it was the only thing depending on extensions/common, so now the core archcomponents only depends on the very lightweight runtime dependency
- Add kotlin extensions artifacts for all three android artifacts

Resolves #87 
Resolves #112 
Resolves #113


hzsweers:
I'll do a pass to update license headers, seems my intelliJ lost the profile :|. Test failure should be easy to fix too

gradle/dependencies.gradle
```diff
@@ -66,7 +66,8 @@ def support = [
         lifecycle: [
             common: "android.arch.lifecycle:common:${versions.archComponents}",
             compiler: "android.arch.lifecycle:compiler:${versions.archComponents}",
-            extensions: "android.arch.lifecycle:extensions:${versions.archComponents}"
+            extensions: "android.arch.lifecycle:extensions:${versions.archComponents}",
+            runtime: "android.arch.lifecycle:runtime:1.0.0"
```
0legg: Do runtime have different release cycle, so you don't use the same version?

settings.gradle
```diff
@@ -15,9 +15,13 @@
  */
 
 rootProject.name = 'autodispose-root'
+include ':android:autodispose-android'
+include ':android:autodispose-android-kotlin'
+include ':android:autodispose-android-archcomponents'
+include ':android:autodispose-android-archcomponents-kotlin'
+include ':android:autodispose-android-archcomponents-test'
+include ':android:autodispose-android-archcomponents-test-kotlin'
```
0legg: Well, I have some concerns here. Looks like we have few dimensions (like lifecycle events provider, platform, language), and implementing or updating new thing for any of it leads to rewriting of N sublibraries. Maybe you have an idea how to resolve it, or we can discuss offline and try to find a solution.

gradle/dependencies.gradle
hzsweers: Yeah runtime is already stable and released

settings.gradle
hzsweers: What do you mean by rewriting?

0legg: Like, imagine, you want to update autodispose-kotlin. It'll (possibly) lead to updates in autodispose-android-archcomponents-kotlin and autodispose-android-archcomponents-test-kotlin (actually, any artifact mentions kotlin).The same condition will stay for, say, update in archcomponents. If we have it somewhat decoupled (I don't have solution for this, it's only an brainstorm-ish idea!), then we can leverage modularity better. Also, it'll be better for library consumers to have more granular control over dependencies.

hzsweers: Ah, no the plan is for them to always be in sync even if some artifacts have no changes

hzsweers:
I can't make heads or tails of this error. All the android artifacts use different, unique package names 🤔

PR #124, opened 2017-10-25
Plugin for controlling whether or not to fill in stacktraces

### LINKED ISSUE #117: Proposal: Plugin point to make fillInStacktrace() configurable for lifecycle boundary exceptions
Exceptions aren't free. Currently we use exceptions to indicate lifecycle boundaries. Normally their trace would be helpful, but in some setups (such as our internal one), users track sources of exceptions in rx chains separately and just use the exception as a signal. Perhaps we can save a little performance by making a plugin hook to configure this. Then when receiving exceptions, traces would not be inspectable and the type would be your only clue.

This would be possible for pre-java8 too by overriding `fillInStacktrace()` and controlling behavior there. The default would be to remain enabled.

```java
public class OutsideLifecycleException extends RuntimeException {

  public OutsideLifecycleException(String s) {
    super(s);
  }

  @Override public final synchronized Throwable fillInStackTrace() {
    if (AutoDisposePlugins.fillInStacktraces()) {
      return super.fillInStackTrace();
    } else {
      return this;
    }
  }
}
```

Questions:
- Should we?
- If no, why not?
- If yes
  - `fillInStacktraces()` method? 
  - Static volatile boolean? (might not play well with `lockdown()`)

hzsweers:
Resolves #117

This implements support for controlling whether or not `OutsideLifecycleException`s fill in their stacktraces. This is a potentially expensive operation that is often used just for signaling, so we can save a little perf here.

### DIFF
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDisposePlugins.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDisposePlugins.java
@@ -28,6 +28,7 @@ public final class AutoDisposePlugins {
 
   @Nullable
   private static volatile Consumer<? super OutsideLifecycleException> outsideLifecycleHandler;
+  private static volatile boolean fillInOutsideLifecycleExceptionStacktraces;
 
   /**
    * Prevents changing the plugins.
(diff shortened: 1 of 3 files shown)


tonycosentini:
Is it possible to add more details about what the specific performance improvements are?

Also +1 to Leland's comment about including example stacktraces to demonstrate the differences.

hzsweers:
Details beyond

> This is a potentially expensive operation that is often used just for signaling, so we can save a little perf here.

?

I’ll add an example, but bear in mind the point is that there is no trace, just the exception name and message. In a conventional “stacktrace”, the difference is that you effictively only get the first line here

tonycosentini:
What is expensive? Just creating the stacktrace itself? Whatever it is, just mentioning it might be useful.

hzsweers:
Yes, `fillInStacktrace()` is (this is detailed in the linked issue as well)

tonycosentini:
Is it worth adding the same details to the javadoc?

hzsweers:
Beyond what's already there?

> Disabling them, if you don't care about the stacktraces, can result in some minor performance improvements.

tonycosentini:
Yes, beyond what's already there.

hzsweers:
(talked and the javadoc is fine as-is)

PR #125, opened 2017-10-31
Move unbound() shorthand to ScopeProvider itself

hzsweers:
This makes `UNBOUND` a singleton instance on `ScopeProvider` itself, allowing for its use in regular non-test code too and a more flexible shorthand.


jbarr21:
+1

Issue #122, opened 2017-10-23
Dokka for kotlin artifacts


hzsweers:
https://github.com/Kotlin/dokka

Kietzmann:
I'm on it

hzsweers:
Thanks! Keep us posted how it goes

hzsweers:
Thanks! Keep us posted how it goes

Hackertronix:
@hzsweers I am trying to give this a shot but I am getting issues importing the project into IntelliJ
![screen shot 2017-10-25 at 11 06 33 am](https://user-images.githubusercontent.com/5685314/31982307-d3252812-b975-11e7-8226-bc6a5c8c264f.png)

Kietzmann:
@Hackertronix 
You must install android sdk

hzsweers:
yep, or just comment out the android modules in `settings.gradle`

Hackertronix:
@kietzmann @hzsweers I already have the SDK installed and configured.... My ANDROID_HOME path is also set up even then i'm getting this

hzsweers:
Even if you just comment out those modules from settings.gradle?

hzsweers:
Or I suppose you might want to have those synced to wire up dokka for the Android javadoc tasks too. I'm not sure other than trying to restart your IDE or to add to to a local.properties file in your project dir

charlesdurham:
@Kietzmann any updates? :)

Kietzmann:
@charlesdurham 
Hi, I faced some issues when building the project. Currently busy and won't have free time for this next week.

charlesdurham:
@Kietzmann would you mind if I looked at the issue in the meantime?

Kietzmann:
No, I wouldn't

2017-11-05 16:05 GMT+02:00 Charles Durham <notifications@github.com>:

hzsweers:
This is done in #126

PR #126, opened 2017-11-05
Add dokka artifacts #122

charlesdurham:
So there's a lot to validate here but I think I've covered my bases, hopefully lol.

A few things I noticed:
- Cross module linking isn't implemented with my dokka tasks, but I don't think it's working with javadoc either. This spits out a fair amount of warnings running the documentation tasks.
- I was able to externally link all the kdoc to rxjava doc, I looked around briefly to try and do the same for the javadoc but didn't find anything that worked; I tried `options.link()`. 
- The source jar for android-kotlin modules was empty in master, I fixed this in 15697057db09d136c1af4b87b673190c4b27a6ce but don't really know why it wasn't working before.

I split out the artifact tasks into their own files because there's 4 different documentation tasks now;
- Android + Kotlin
- Android + Java
- Kotlin
- Java

I like this more than adding conditionals to `gradle-mvn-push.gradle` but I'm not super strongly opinionated so if you feel like it should all go into that file let me know and I'll update.


gradle/dependencies.gradle
```diff
@@ -41,7 +42,9 @@ def build = [
         android: 'com.android.tools.build:gradle:2.3.0',
         apt: 'net.ltgt.gradle:gradle-apt-plugin:0.11',
         errorProne: 'net.ltgt.gradle:gradle-errorprone-plugin:0.0.11',
-        kotlin: "org.jetbrains.kotlin:kotlin-gradle-plugin:${versions.kotlin}"
+        kotlin: "org.jetbrains.kotlin:kotlin-gradle-plugin:${versions.kotlin}",
+        dokka: "org.jetbrains.dokka:dokka-gradle-plugin:${versions.dokka}",
```
hzsweers: Nit: alpha this

gradle/gradle-mvn-push.gradle
```diff
@@ -42,6 +42,101 @@ def getRepositoryPassword() {
...
 }
 
+def kotlinAndroidArtifactTasks() {
+  if (!project.plugins.hasPlugin('org.jetbrains.dokka-android')) {
+    throw new GradleException("Apply the dokka-android plugin in ${project.name}")
+  }
+
+  dokka {
+    externalDocumentationLink {
+      url = new URL("http://reactivex.io/RxJava/2.x/javadoc/")
```
hzsweers: should this point to https://uber.github.io/AutoDispose/0.x/ ?

gradle/gradle-mvn-push.gradle
```diff
@@ -42,6 +42,101 @@ def getRepositoryPassword() {
...
+}
+
+def kotlinArtifactTasks() {
+  if (!project.plugins.hasPlugin('org.jetbrains.dokka')) {
+    throw new GradleException("Apply the dokka plugin in ${project.name}")
+  }
+
+  dokka {
+    externalDocumentationLink {
+      url = new URL("http://reactivex.io/RxJava/2.x/javadoc/")
```
hzsweers: Same here

gradle/gradle-mvn-push.gradle
charlesdurham: Hm, what are you thinking here? That's what builds in linking between the kdoc and rxjava's doc, so if you're looking at documentation that links to rxjava you can click on it and jump straight to the rxjava documentation. 

I could see adding https://uber.github.io/AutoDispose/0.x/ being useful (at least until there's proper inter-module linking). Although I think each module's doc might need to be linked independently. The dokka examples indicate you can have multiple external documentation links so if you want I could add it and keep rxjava.

I actually didn't even realize the documentation was hosted at  https://uber.github.io/AutoDispose/0.x/, this PR just creates javadoc artifacts for maven. 

I see https://github.com/uber/AutoDispose/tree/gh-pages/0.x but don't really know how/if the process of getting the doc in there is automated. If you can point me in the right direction I'd be happy to include that as part of this PR.

hzsweers: yeah I'd held off on publicizing the docs site till dokka was figured out. I upload them via [osstrich](https://github.com/square/osstrich), unfortunately not sure automating it is viable vs just doing it manually after releases (don't worry about it for this PR). I didn't realize that was for dependencies of the project though, so this makes sense :+1:

hzsweers:
Thanks! I'll give this a spin at the next release, which should be soon-ish

PR #127, opened 2017-11-07
Add subscribeBy extensions for proxy classes.

ajalt:
As we discussed at Droidcon, since the SubscribeProxy classes don't extend the
observable class they proxy to, extension functions written for the
observable class can't be used.

One commonly used set of extensions is RxKotlin's [subscribeBy](https://github.com/ReactiveX/RxKotlin/blob/2.x/src/main/kotlin/io/reactivex/rxkotlin/subscribers.kt). This
commit adds copies of those extensions that work on SubscribeProxies.


ajalt:
I think that keeping the artifacts small is a generally good idea, and these extensions could be considered outside the scope of the library. It seems like these particular extensions might be worth including due to the fact that they are the only ones in RxKotlin that don't work with AutoDispose, and they are fairly widely used. But it's up to you.

hzsweers:
My gut says it's better to make these a separate thing - erring on the side of restricting the API surface area in autodispose itself. If you want to put them in the sample though, I'd be happy to merge that and point to them as a recipe for RxKotlin users. This also sounds like a good library idea for you if you wanted to maintain it yourself :)

ajalt:
Ok, I moved the extensions to the sample recipes and updated the activity to use them. Is that what you had in mind?

sample/src/main/kotlin/com/uber/autodispose/recipes/subscriberproxies.kt
```diff
@@ -0,0 +1,54 @@
+package com.uber.autodispose.recipes
+
+import com.uber.autodispose.*
```
hzsweers: nit: no wildcards

sample/src/main/kotlin/com/uber/autodispose/recipes/subscriberproxies.kt
```diff
@@ -0,0 +1,54 @@
...
+import io.reactivex.disposables.Disposable
+import io.reactivex.exceptions.OnErrorNotImplementedException
+import io.reactivex.plugins.RxJavaPlugins
+
+private val onNextStub: (Any) -> Unit = {}
+private val onErrorStub: (Throwable) -> Unit = { RxJavaPlugins.onError(OnErrorNotImplementedException(it)) }
+private val onCompleteStub: () -> Unit = {}
+
+
```
hzsweers: nit: single blank line

sample/src/main/kotlin/com/uber/autodispose/recipes/subscriberproxies.kt
```diff
@@ -0,0 +1,54 @@
+package com.uber.autodispose.recipes
+
+import com.uber.autodispose.*
+import io.reactivex.disposables.Disposable
+import io.reactivex.exceptions.OnErrorNotImplementedException
+import io.reactivex.plugins.RxJavaPlugins
+
+private val onNextStub: (Any) -> Unit = {}
```
hzsweers: Let's add a file doc above this that explains how this is an example of how you can tack on extension functions to the proxy interface

```kotlin
/*
 * Stuff!
 */
```

sample/src/main/kotlin/com/uber/autodispose/recipes/subscriberproxies.kt
```diff
@@ -0,0 +1,54 @@
...
+private val onErrorStub: (Throwable) -> Unit = { RxJavaPlugins.onError(OnErrorNotImplementedException(it)) }
```
hzsweers: I'd like to propose diverging from how RxKotlin does this. The reason is that RxJava actually does special handling for missing error implementations via [`LambdaConsumerIntrospection`](https://github.com/ReactiveX/RxJava/pull/5590), that supplying this would not catch. I'd say for the cases where only a next/success/complete consumer/action is passed in, let's call through to the unhandled one in rxjava. Will comment inline with specifics

sample/src/main/kotlin/com/uber/autodispose/recipes/subscriberproxies.kt
```diff
@@ -0,0 +1,54 @@
...
+
+
+/**
+ * Overloaded subscribe function that allows passing named parameters
+ */
+fun <T : Any> ObservableSubscribeProxy<T>.subscribeBy(
+        onError: (Throwable) -> Unit = onErrorStub,
+        onComplete: () -> Unit = onCompleteStub,
+        onNext: (T) -> Unit = onNextStub
+): Disposable = subscribe(onNext, onError, onComplete)
```
hzsweers: Adjust for `LambdaConsumerIntrospection`

```kotlin
if (onError === onErrorStub && onComplete === onCompleteStub) {
  return subscribe(onNext)
} else {
  return subscribe(onNext, onError, onComplete)
}
```

sample/src/main/kotlin/com/uber/autodispose/recipes/subscriberproxies.kt
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Overloaded subscribe function that allows passing named parameters
+ */
+fun <T : Any> FlowableSubscribeProxy<T>.subscribeBy(
+        onError: (Throwable) -> Unit = onErrorStub,
+        onComplete: () -> Unit = onCompleteStub,
+        onNext: (T) -> Unit = onNextStub
+): Disposable = subscribe(onNext, onError, onComplete)
```
hzsweers: Adjust for `LambdaConsumerIntrospection`

```kotlin
if (onError === onErrorStub && onComplete === onCompleteStub) {
 return subscribe(onNext)
} else {
 return subscribe(onNext, onError, onComplete)
}
```

sample/src/main/kotlin/com/uber/autodispose/recipes/subscriberproxies.kt
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Overloaded subscribe function that allows passing named parameters
+ */
+fun <T : Any> SingleSubscribeProxy<T>.subscribeBy(
+        onError: (Throwable) -> Unit = onErrorStub,
+        onSuccess: (T) -> Unit = onNextStub
+): Disposable = subscribe(onSuccess, onError)
```
hzsweers: Adjust for `LambdaConsumerIntrospection`

```kotlin
if (onError === onErrorStub) {
 return subscribe(onSuccess)
} else {
 return subscribe(onSuccess, onError)
}
```

sample/src/main/kotlin/com/uber/autodispose/recipes/subscriberproxies.kt
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Overloaded subscribe function that allows passing named parameters
+ */
+fun <T : Any> MaybeSubscribeProxy<T>.subscribeBy(
+        onError: (Throwable) -> Unit = onErrorStub,
+        onComplete: () -> Unit = onCompleteStub,
+        onSuccess: (T) -> Unit = onNextStub
+): Disposable = subscribe(onSuccess, onError, onComplete)
```
hzsweers: Adjust for `LambdaConsumerIntrospection`

```kotlin
if (onError === onErrorStub && onComplete === onCompleteStub) {
 return subscribe(onSuccess)
} else {
 return subscribe(onSuccess, onError, onComplete)
}
```

sample/src/main/kotlin/com/uber/autodispose/recipes/subscriberproxies.kt
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Overloaded subscribe function that allows passing named parameters
+ */
+fun CompletableSubscribeProxy.subscribeBy(
+        onError: (Throwable) -> Unit = onErrorStub,
+        onComplete: () -> Unit = onCompleteStub
+): Disposable = subscribe(onComplete, onError)
```
hzsweers: Adjust for `LambdaConsumerIntrospection`

```kotlin
if (onError === onErrorStub) {
 return subscribe(onComplete)
} else {
 return subscribe(onComplete, onError)
}
```

sample/src/main/kotlin/com/uber/autodispose/sample/KotlinActivity.kt
```diff
@@ -45,7 +46,7 @@ class KotlinActivity : AppCompatActivity() {
     Observable.interval(1, TimeUnit.SECONDS)
         .doOnDispose { Log.i(TAG, "Disposing subscription from onCreate()") }
         .autoDisposeWith(scopeProvider)
-        .subscribe { num -> Log.i(TAG, "Started in onCreate(), running until onDestroy(): " + num) }
```
hzsweers: nit: let's keep the `num`, as it was an intentionally more specific name

sample/src/main/kotlin/com/uber/autodispose/recipes/subscriberproxies.kt
ajalt: Hm, interesting. I added RxKotlin's [current implementation](https://github.com/ReactiveX/RxKotlin/pull/141) before `LambdaConsumerIntrospection` was in place, and it matched RxJava's behavior at the time. I'll look into your PR in detail later, but do you think RxKotlin should also be updated?

hzsweers: Yeah I'd say it probably should. Would be a non-breaking change, but it would allow rxjava to properly expose the introspection details

ajalt:
I'm not sure I understand what you mean by filling out method bodies?

hzsweers:
As in instead of:

```kotlin
fun <T : Any> ObservableSubscribeProxy<T>.subscribeBy(
        onError: (Throwable) -> Unit = onErrorStub,
        onComplete: () -> Unit = onCompleteStub,
        onNext: (T) -> Unit = onNextStub
): Disposable =
        if (onError === onErrorStub && onComplete === onCompleteStub) {
            subscribe(onNext)
        } else {
            subscribe(onNext, onError, onComplete)
        }
```

Do this

```kotlin
fun <T : Any> ObservableSubscribeProxy<T>.subscribeBy(
        onError: (Throwable) -> Unit = onErrorStub,
        onComplete: () -> Unit = onCompleteStub,
        onNext: (T) -> Unit = onNextStub
): Disposable {
  return if (onError === onErrorStub && onComplete === onCompleteStub) {
    subscribe(onNext)
  } else {
    subscribe(onNext, onError, onComplete)
  }
}
```

PR #135, opened 2017-11-15
Fix a racing condition when subscribing to a completed ScopeProvider.

duanbo1983:
**Description**:

**Related issue(s)**:

### DIFF
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDisposingObserverImpl.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDisposingObserverImpl.java
@@ -46,11 +46,14 @@ final class AutoDisposingObserverImpl<T> implements AutoDisposingObserver<T> {
           @Override public void onSuccess(Object o) {
             callMainSubscribeIfNecessary(d);
             AutoDisposingObserverImpl.this.dispose();
+            lifecycleDisposable.lazySet(AutoDisposableHelper.DISPOSED);
           }
 
           @Override public void onError(Throwable e) {
             callMainSubscribeIfNecessary(d);
             AutoDisposingObserverImpl.this.onError(e);
+            lifecycleDisposable.lazySet(AutoDisposableHelper.DISPOSED);
+            mainDisposable.lazySet(AutoDisposableHelper.DISPOSED);
           }
 
           @Override public void onComplete() {
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDisposingObserverImpl.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDisposingObserverImpl.java
@@ -55,6 +55,8 @@ final class AutoDisposingObserverImpl<T> implements AutoDisposingObserver<T> {
 
           @Override public void onComplete() {
             callMainSubscribeIfNecessary(d);
+            lifecycleDisposable.lazySet(AutoDisposableHelper.DISPOSED);
+            mainDisposable.lazySet(AutoDisposableHelper.DISPOSED);
             // Noop - we're unbound now
           }
         }),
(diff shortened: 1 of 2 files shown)


duanbo1983:
Let me switch an account.

autodispose/src/main/java/com/uber/autodispose/AutoDisposingObserverImpl.java
```diff
@@ -55,6 +55,8 @@
 
           @Override public void onComplete() {
             callMainSubscribeIfNecessary(d);
```
duanbo1983: When subscribing to a completed ScopeProvider, callMainSubscribeIfNecessary() might be called here before the main source is subscribed, which triggers the DoubleSubscriptoinsException.

autodispose/src/test/java/com/uber/autodispose/AutoDisposeObserverTest.java
```diff
@@ -171,6 +177,24 @@
     assertThat(lifecycle.hasObservers()).isFalse();
   }
 
+  @Test public void autoDispose_withScopeProviderCompleted_shouldNotReportDoubleSubscriptions() {
+    RxJavaPlugins.setErrorHandler(new Consumer<Throwable>() {
+      @Override
+      public void accept(Throwable throwable) throws Exception {
+        assertThat(throwable instanceof ProtocolViolationException).isFalse();
```
duanbo1983: This test case fails in master as:
`Exception in thread "main" io.reactivex.exceptions.ProtocolViolationException: It is not allowed to subscribe with a(n) com.uber.autodispose.AutoDisposingObserverImpl multiple times. Please create a fresh instance of com.uber.autodispose.AutoDisposingObserverImpl and subscribe that to the target source instead.
	at com.uber.autodispose.AutoDisposeEndConsumerHelper.reportDoubleSubscription(AutoDisposeEndConsumerHelper.java:110)
	at com.uber.autodispose.AutoDisposeEndConsumerHelper.setOnce(AutoDisposeEndConsumerHelper.java:56)
	at com.uber.autodispose.AutoDisposingObserverImpl.onSubscribe(AutoDisposingObserverImpl.java:67)
	at io.reactivex.subjects.PublishSubject.subscribeActual(PublishSubject.java:85)
	at io.reactivex.Observable.subscribe(Observable.java:10842)
	at com.uber.autodispose.ObservableScoper$AutoDisposeObservable.subscribeActual(ObservableScoper.java:113)
	at io.reactivex.Observable.subscribe(Observable.java:10842)
	at com.uber.autodispose.ObservableScoper$1.subscribe(ObservableScoper.java:94)
	at com.uber.autodispose.AutoDisposeObserverTest.autoDispose_withScopeProviderCompleted_shouldNotReportDoubleSubscriptions(AutoDisposeObserverTest.java:192)`

autodispose/src/main/java/com/uber/autodispose/AutoDisposingObserverImpl.java
```diff
@@ -46,15 +46,20 @@
           @Override public void onSuccess(Object o) {
             callMainSubscribeIfNecessary(d);
             AutoDisposingObserverImpl.this.dispose();
```
hzsweers: This should be changed to `AutoDisposableHelper.dispose(mainDisposable)`

autodispose/src/test/java/com/uber/autodispose/AutoDisposeObserverTest.java
hzsweers: I don't quite understand this test, as the `ProtocolViolationException` isn't obvious. Some sort of comment and asserting what you do expect should happen here

It also won't fail if this callback isn't hit, let's make it explicit somehow that we expect this callback to be hit

ZacSweers (changes requested):
Good start. Can you also update the other observers? i.e. for `Single`, `Maybe`, `Completable`, and `Flowable`

hzsweers:
Talked offline, closing in favor of #138

Issue #130, opened 2017-11-10
Review of RxJava 2 usage in the project


akarnokd:
Hi. I was asked to review the library's RxJava 2 usage for potential logical or concurrency issues.

Review @ commit: [0147484b](https://github.com/uber/AutoDispose/commit/0147484bdc6237e48bb1038faef3dd4e743025bf)

### Main issues:

- [AutoDisposingCompletableObserverImpl.java#L45](https://github.com/uber/AutoDispose/blob/master/autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java#L45): The `DisposableMaybeObserver` could be created and `setOnce`'d before the call to `lifecycle.subscribe()`.
- [AutoDisposingCompletableObserverImpl.java#L52](https://github.com/uber/AutoDispose/blob/master/autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java#L52): The `mainDisposable` could be already set and the `delegate` currently executing one of its `onXXX` methods which violates the contract unless it is guaranteed both the lifecycle signal and the onXXX calls happen on the same thread (i.e., forced `observeOn(AndroidSchedulers.mainThread())`). Possible solution: always call `delegate.onSubscribe(this)` first, mutually exclude the onXXX signals via an `AtomicBoolean once`.
- [AutoDisposingCompletableObserverImpl.java#L58](https://github.com/uber/AutoDispose/blob/0147484bdc6237e48bb1038faef3dd4e743025bf/autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java#L58): May not be a good idea to leave the delegate non-terminated in this case, if the lifecycle actually can get `onComplete`.
- [AutoDisposingCompletableObserverImpl.java#L73](https://github.com/uber/AutoDispose/blob/master/autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java#L73): `synchronized` is pointless here.
- [AutoDisposingCompletableObserverImpl.java#L80](https://github.com/uber/AutoDispose/blob/master/autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java#L80): `synchronized` is pointless here. If the purpose was to mutually exclude with `#L73` then it is not worth it. If `lazySet` wins then `dispose(ref)` will do nothing, if they race, both cases could win probabilistically.
[...]

PR #138, opened 2017-11-18
Incorporate feedback from #130

hzsweers:
This PR incorporates feedback from #130, but it not currently complete yet. CC @akarnokd whenever you have time to review (no rush!)

## Completed

> AutoDisposePlugins.java#L36: Such lockdown support may not be desirable on Android. The same feature was aimed at server/J2EE container use of RxJava.

Mentioned in the issue, opted to keep this to match semantics of RxJava's plugins. We also operate on a large monorepo structure internally where this actually becomes useful to help catch downstream usages.

> dependencies.gradle#L65: Using a pretty old RxJava 2 version.

Touched on in the issue, and after thinking on it I think I'm going to keep this as is until there's a new minor release or new API we want to target. Keeping the lower API makes it easier for consumers to adopt without imposing newer versions if they haven't been able to update yet.

> AutoDisposingCompletableObserverImpl.java#L73: synchronized is pointless here.
> AutoDisposingCompletableObserverImpl.java#L80: synchronized is pointless here. If the purpose was to mutually exclude with #L73 then it is not worth it. If lazySet wins then dispose(ref) will do nothing, if they race, both cases could win probabilistically.

bd22bc2

> AutoDisposingCompletableObserver.java#L30: missing space in forintrospection.

caf904a

> AutoDisposingObserver.java#L30: missing tailing s from purpose?

79e6dbf

> AutoDispose.java#L169: class could be final.
> AutoDispose.java#L199: class could be final.
> AutoDispose.java#L229: class could be final.

94212eb

> AutoDisposingCompletableObserverImpl.java#L45: The DisposableMaybeObserver could be created and setOnce'd before the call to lifecycle.subscribe().

9bf1993
 
> AutoDisposingObserverImpl.java: Same problems as with the CompletableObserver implementation. Here one would need a HalfSerializer to mutually exclude the onXXX events (as there could be multiple onNext calls).

cb432a8. I think I've mostly got this. Main thing I'm not sure of is if it does terminate, do we need to dispose the lifecycle or does this mean that onError or onComplete have been hit in the main observer already (and doing so is a double dispose condition)

> ViewAttachEventsObservable.java#L41 If the call is not on the main thread, onSubscribe is not called which can lead to a NullPointerException in the dowstream. The no-op Disposable should be sent to the Observer first.

07f0f1e and corresponding one for `LifecycleEventsObservable` ec190c0

> ViewAttachEventsObservable.java#L51: The Observer may cancel immediately and since the listener is not registered with the view yet, it won't get removed but then added unconditionally. Assuming the events also come from the main thread, one could add the listener first, and call onSubscribe next. If adding the listener triggers an event emission immediately, keep the current order and after adding the listener, check if the listener has been disposed in the meantime and remove the listener.

I think this is covered as a part of 07f0f1e, but let me know if I've missed something

> AutoDisposingSubscriberImpl.java#L80: There is no onStart provided by the API and the delegate.onSubscribe is called after mainSubscription is definitely set: the null check should always pass and thus unnecessary.

b38d03f

## TODO

> AutoDisposingCompletableObserverImpl.java#L52: The mainDisposable could be already set and the delegate currently executing one of its onXXX methods which violates the contract unless it is guaranteed both the lifecycle signal and the onXXX calls happen on the same thread (i.e., forced observeOn(AndroidSchedulers.mainThread())). Possible solution: always call delegate.onSubscribe(this) first, mutually exclude the onXXX signals via an AtomicBoolean once.

If I subscribe first - couldn't it synchronously execute its entire sequence before the lifecycle gets a chance then? Part of ths reason it's done ater the lifecycle check is such that the lifecycle can indicate that it's ended, and just immediately dispose after onSubscribe().

> AutoDisposingCompletableObserverImpl.java#L58: May not be a good idea to leave the delegate non-terminated in this case, if the lifecycle actually can get onComplete.

Could you expand on this one a bit? I'm not sure I totally understood

> AutoDisposingSubscriberImpl.java#L80: Note though that if one does the setOnce(mainSubscription) before the lifecycle is attached, one may need to use deferred requesting to not trigger the upstream emission while the code is setting up other things in onSubscribe.

See comment below

### DIFF
--- a/android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/LifecycleEventsObservable.java
+++ b/android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/LifecycleEventsObservable.java
@@ -84,6 +84,9 @@ import static com.uber.autodispose.android.internal.AutoDisposeAndroidUtil.isMai
       return;
     }
     lifecycle.addObserver(archObserver);
+    if (archObserver.isDisposed()) {
+      lifecycle.removeObserver(archObserver);
+    }
   }
 
   static final class ArchLifecycleObserver extends MainThreadDisposable
--- a/android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/LifecycleEventsObservable.java
+++ b/android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/LifecycleEventsObservable.java
@@ -75,14 +75,14 @@ import static com.uber.autodispose.android.internal.AutoDisposeAndroidUtil.isMai
   }
 
   @Override protected void subscribeActual(Observer<? super Event> observer) {
+    ArchLifecycleObserver archObserver =
+        new ArchLifecycleObserver(lifecycle, observer, eventsObservable);
+    observer.onSubscribe(archObserver);
     if (!isMainThread()) {
       observer.onError(
           new IllegalStateException("Lifecycles can only be bound to on the main thread!"));
       return;
     }
-    ArchLifecycleObserver archObserver =
-        new ArchLifecycleObserver(lifecycle, observer, eventsObservable);
-    observer.onSubscribe(archObserver);
     lifecycle.addObserver(archObserver);
   }
 
--- a/android/autodispose-android/src/main/java/com/uber/autodispose/android/ViewAttachEventsObservable.java
+++ b/android/autodispose-android/src/main/java/com/uber/autodispose/android/ViewAttachEventsObservable.java
@@ -37,6 +37,8 @@ final class ViewAttachEventsObservable extends Observable<ViewLifecycleEvent> {
   }
 
   @Override protected void subscribeActual(Observer<? super ViewLifecycleEvent> observer) {
+    Listener listener = new Listener(view, observer);
+    observer.onSubscribe(listener);
     if (!isMainThread()) {
       observer.onError(new IllegalStateException("Views can only be bound to on the main thread!"));
       return;
@@ -46,8 +48,6 @@ final class ViewAttachEventsObservable extends Observable<ViewLifecycleEvent> {
       // Emit the last event, like a behavior subject
       observer.onNext(ViewLifecycleEvent.ATTACH);
     }
-    Listener listener = new Listener(view, observer);
-    observer.onSubscribe(listener);
     view.addOnAttachStateChangeListener(listener);
   }
 
--- a/android/autodispose-android/src/main/java/com/uber/autodispose/android/ViewAttachEventsObservable.java
+++ b/android/autodispose-android/src/main/java/com/uber/autodispose/android/ViewAttachEventsObservable.java
@@ -49,6 +49,9 @@ final class ViewAttachEventsObservable extends Observable<ViewLifecycleEvent> {
       observer.onNext(ViewLifecycleEvent.ATTACH);
     }
     view.addOnAttachStateChangeListener(listener);
+    if (listener.isDisposed()) {
+      view.removeOnAttachStateChangeListener(listener);
+    }
   }
 
   static final class Listener extends MainThreadDisposable
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDispose.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDispose.java
@@ -166,7 +166,7 @@ public final class AutoDispose {
     return new LifecycleScopeProviderHandlerImpl(scope);
   }
 
-  private static class MaybeScopeHandlerImpl implements ScopeHandler {
+  private static final class MaybeScopeHandlerImpl implements ScopeHandler {
 
     private final Maybe<?> scope;
 
@@ -196,7 +196,7 @@ public final class AutoDispose {
     }
   }
 
-  private static class ScopeProviderHandlerImpl implements ScopeHandler {
+  private static final class ScopeProviderHandlerImpl implements ScopeHandler {
 
     private final ScopeProvider scope;
 
@@ -226,7 +226,7 @@ public final class AutoDispose {
     }
   }
 
-  private static class LifecycleScopeProviderHandlerImpl implements ScopeHandler {
+  private static final class LifecycleScopeProviderHandlerImpl implements ScopeHandler {
 
     private final LifecycleScopeProvider<?> scope;
 
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDispose.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDispose.java
@@ -168,7 +168,7 @@ public final class AutoDispose {
 
   private static final class MaybeScopeHandlerImpl implements ScopeHandler {
 
-    private final Maybe<?> scope;
+    final Maybe<?> scope;
 
     MaybeScopeHandlerImpl(Maybe<?> scope) {
       this.scope = scope;
@@ -198,7 +198,7 @@ public final class AutoDispose {
 
   private static final class ScopeProviderHandlerImpl implements ScopeHandler {
 
-    private final ScopeProvider scope;
+    final ScopeProvider scope;
 
     ScopeProviderHandlerImpl(ScopeProvider scope) {
       this.scope = scope;
@@ -228,7 +228,7 @@ public final class AutoDispose {
 
   private static final class LifecycleScopeProviderHandlerImpl implements ScopeHandler {
 
-    private final LifecycleScopeProvider<?> scope;
+    final LifecycleScopeProvider<?> scope;
 
     LifecycleScopeProviderHandlerImpl(LifecycleScopeProvider<?> scope) {
       this.scope = scope;
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java
@@ -79,11 +79,6 @@ final class AutoDisposingCompletableObserverImpl implements AutoDisposingComplet
     AutoDisposableHelper.dispose(mainDisposable);
   }
 
-  private void lazyDispose() {
-    AutoDisposableHelper.dispose(lifecycleDisposable);
-    mainDisposable.lazySet(AutoDisposableHelper.DISPOSED);
-  }
-
   @SuppressWarnings("WeakerAccess") // Avoiding synthetic accessors
   void callMainSubscribeIfNecessary(Disposable d) {
     // If we've never actually called the downstream onSubscribe (i.e. requested immediately in
@@ -96,14 +91,16 @@ final class AutoDisposingCompletableObserverImpl implements AutoDisposingComplet
 
   @Override public void onComplete() {
     if (!isDisposed()) {
-      lazyDispose();
+      AutoDisposableHelper.dispose(lifecycleDisposable);
+      mainDisposable.lazySet(AutoDisposableHelper.DISPOSED);
       delegate.onComplete();
     }
   }
 
   @Override public void onError(Throwable e) {
     if (!isDisposed()) {
-      lazyDispose();
+      AutoDisposableHelper.dispose(lifecycleDisposable);
+      mainDisposable.lazySet(AutoDisposableHelper.DISPOSED);
       delegate.onError(e);
     }
   }
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java
@@ -47,9 +47,8 @@ final class AutoDisposingCompletableObserverImpl implements AutoDisposingComplet
       }
 
       @Override public void onError(Throwable e) {
-        AutoDisposingCompletableObserverImpl.this.onError(e);
-        mainDisposable.lazySet(AutoDisposableHelper.DISPOSED);
         lifecycleDisposable.lazySet(AutoDisposableHelper.DISPOSED);
+        AutoDisposingCompletableObserverImpl.this.onError(e);
       }
 
       @Override public void onComplete() {
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java
@@ -70,17 +70,13 @@ final class AutoDisposingCompletableObserverImpl implements AutoDisposingComplet
   }
 
   @Override public void dispose() {
-    synchronized (this) {
-      AutoDisposableHelper.dispose(lifecycleDisposable);
-      AutoDisposableHelper.dispose(mainDisposable);
-    }
+    AutoDisposableHelper.dispose(lifecycleDisposable);
+    AutoDisposableHelper.dispose(mainDisposable);
   }
 
   private void lazyDispose() {
-    synchronized (this) {
-      AutoDisposableHelper.dispose(lifecycleDisposable);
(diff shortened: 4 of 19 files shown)


hzsweers:
> Could you expand on this one a bit? I haven't used flowable's APIs that much so I'm a little inexperienced with handling requesting or deferring it

I actually missed the link in your original comment. Going to research more with that and report back!

hzsweers:
Ok I think I've got deferred requesting set up in 500f337 after reading the docs. If that looks good, I'll bring those methods into the `AutoDisposeSubscriptionHelper` copy so it's not depending directly on internal classes

hzsweers:
Ok, I'm missing something about how to wire up deferred requesting. Let me know if you have any advice!

akarnokd:
Here is the [wiki](https://github.com/ReactiveX/RxJava/wiki/Writing-operators-for-2.0#deferred-requesting) entry for deferred requesting.

>> AutoDisposingCompletableObserverImpl.java#L58: May not be a good idea to leave the delegate non-terminated in this case, if the lifecycle actually can get onComplete.

> Could you expand on this one a bit? I'm not sure I totally understood

Does the lifecycle's `onComplete` get called while there is still a subscriber active? Consider this:

```java
PublishSubject.create()
.to(AutoDispose.with(ScopeProvider.UNBOUND).forObservable())
.subscribe(new DisposableObserver<Object>() {
    PrintWriter fout;

    @Override public void onStart() {
        fout = new PrintWriter(new FileWriter("output.txt"));
    }

    @Override public void onNext(Object o) {
        fout.println(o);
    }

    @Override public void onError(Throwable t) {
        t.printStackTrace(fout);
        fout.close();
    }

    @Override public void onComplete() {
        fout.close();
    }
});
```

If the lifecycle ends, `fout` may left open.

hzsweers:
Yep that's correct behavior. Something may be considered "unbound" in the sense that they won't be disposed by the lifecycle ending. In the example case, it would be as if it was running without autodisposing. The intention for `ScopeProvider.UNBOUND` is to be a shortcut for upstream providers to pass down. Doing it exactly as your snippet does would be considered sort of wasteful, in the same way `takeUntil(Observable.just(1))` might be.

For the deferred requesting, I tried implementing pretty much like the `Operator` example in the wiki, but I get an error I don't quite understand when I try to test with it:

```
java.lang.AssertionError: Not subscribed! (latch = 1, values = 0, errors = 0, completions = 0)

	at io.reactivex.observers.BaseTestConsumer.fail(BaseTestConsumer.java:162)
	at io.reactivex.subscribers.TestSubscriber.assertSubscribed(TestSubscriber.java:309)
	at com.uber.autodispose.AutoDisposeSubscriberTest.autoDispose_withProvider(AutoDisposeSubscriberTest.java:124)
```

Does anything stand out to you as off in my implementation?

akarnokd:
Could you point me to the implementation that is targeted by that test?

hzsweers:
Changes -
 https://github.com/uber/AutoDispose/pull/138/commits/500f3373ebc891cf74512e15e9671393a539307f

And the whole impl - https://github.com/uber/AutoDispose/blob/500f3373ebc891cf74512e15e9671393a539307f/autodispose/src/main/java/com/uber/autodispose/AutoDisposingSubscriberImpl.java

akarnokd:
You have to call `delegate.onSubscribe(this)` before [L74](https://github.com/uber/AutoDispose/blob/500f3373ebc891cf74512e15e9671393a539307f/autodispose/src/main/java/com/uber/autodispose/AutoDisposingSubscriberImpl.java#L74).

hzsweers:
That unfortunately causes several other tests to fail with a double subscription error:

```
java.lang.AssertionError: Error(s) present: [java.lang.IllegalStateException: onSubscribe received multiple subscriptions: EmptySubscription] (latch = 1, values = 0, errors = 1, completions = 0)
	at io.reactivex.observers.BaseTestConsumer.fail(BaseTestConsumer.java:162)
	at io.reactivex.observers.BaseTestConsumer.assertNoErrors(BaseTestConsumer.java:247)
	at com.uber.autodispose.AutoDisposeSubscriberTest.autoDispose_withScopeProviderCompleted_shouldNotReportDoubleSubscriptions(AutoDisposeSubscriberTest.java:347)

...
java.lang.AssertionError: Error(s) present: [java.lang.IllegalStateException: onSubscribe received multiple subscriptions: EmptySubscription] (latch = 1, values = 0, errors = 1, completions = 0)
	at io.reactivex.observers.BaseTestConsumer.fail(BaseTestConsumer.java:162)
	at io.reactivex.observers.BaseTestConsumer.assertNoErrors(BaseTestConsumer.java:247)
	at com.uber.autodispose.AutoDisposeSubscriberTest.autoDispose_withProviderAndNoOpPlugin_afterEnding_shouldFailSilently(AutoDisposeSubscriberTest.java:247)
```

hzsweers:
Also - calling delegate.onSubscribe() before the lifecycle kind of gets to the other remaining concern I had in the description:

> If I subscribe first - couldn't it synchronously execute its entire sequence before the lifecycle gets a chance then? Part of the reason it's done after the lifecycle check is such that the lifecycle can indicate that it's ended, and just immediately dispose after onSubscribe().

Would be curious to know your thoughts

akarnokd:
You don't need to call `callMainSubscribeIfNecessary` as the `delegate`'s `onSubscribe` will be called before the lifecycle is subscribed.

hzsweers:
D'oh, I totally missed that. Fixed in e7e1b87. With that, I think I've covered the points from your issue if you want to do a full review?

autodispose/src/main/java/com/uber/autodispose/AutoDispose.java
```diff
@@ -166,7 +166,7 @@
     return new LifecycleScopeProviderHandlerImpl(scope);
   }
 
-  private static class MaybeScopeHandlerImpl implements ScopeHandler {
+  private static final class MaybeScopeHandlerImpl implements ScopeHandler {
```
akarnokd: I'd avoid `private` in case you don't want to get accessor methods on Android.

autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java
```diff
@@ -41,24 +41,29 @@
...
+        callMainSubscribeIfNecessary(d);
+        mainDisposable.lazySet(AutoDisposableHelper.DISPOSED);
+        lifecycleDisposable.lazySet(AutoDisposableHelper.DISPOSED);
+        // Noop - we're unbound now
+      }
+    };
+    if (AutoDisposeEndConsumerHelper.setOnce(lifecycleDisposable, o, getClass())) {
+      lifecycle.subscribe(o);
       if (AutoDisposeEndConsumerHelper.setOnce(mainDisposable, d, getClass())) {
         delegate.onSubscribe(this);
```
akarnokd: `callMainSubscribeIfNecessary` may be a problem here too causing double `onSubscribe`s.

autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java
```diff
@@ -95,14 +91,16 @@ void callMainSubscribeIfNecessary(Disposable d) {
...
+      mainDisposable.lazySet(AutoDisposableHelper.DISPOSED);
       delegate.onComplete();
     }
   }
 
   @Override public void onError(Throwable e) {
     if (!isDisposed()) {
-      lazyDispose();
+      AutoDisposableHelper.dispose(lifecycleDisposable);
+      mainDisposable.lazySet(AutoDisposableHelper.DISPOSED);
```
akarnokd: Swap these so the `mainDisposable` change gets flushed with the `dispose()` call.

autodispose/src/main/java/com/uber/autodispose/AutoDisposingMaybeObserverImpl.java
```diff
@@ -41,24 +41,29 @@
...
+        callMainSubscribeIfNecessary(d);
+        mainDisposable.lazySet(AutoDisposableHelper.DISPOSED);
+        lifecycleDisposable.lazySet(AutoDisposableHelper.DISPOSED);
+        // Noop - we're unbound now
+      }
+    };
+    if (AutoDisposeEndConsumerHelper.setOnce(lifecycleDisposable, o, getClass())) {
+      lifecycle.subscribe(o);
       if (AutoDisposeEndConsumerHelper.setOnce(mainDisposable, d, getClass())) {
         delegate.onSubscribe(this);
```
akarnokd: `callMainSubscribeIfNecessary` may cause double `onSubscribe` here.

autodispose/src/main/java/com/uber/autodispose/AutoDisposingMaybeObserverImpl.java
```diff
@@ -95,21 +91,24 @@ void callMainSubscribeIfNecessary(Disposable d) {
 
   @Override public void onSuccess(T value) {
     if (!isDisposed()) {
-      lazyDispose();
+      AutoDisposableHelper.dispose(lifecycleDisposable);
+      mainDisposable.lazySet(AutoDisposableHelper.DISPOSED);
```
akarnokd: Swap these.

autodispose/src/main/java/com/uber/autodispose/AutoDisposingObserverImpl.java
```diff
@@ -41,24 +43,29 @@
...
+        callMainSubscribeIfNecessary(d);
+        mainDisposable.lazySet(AutoDisposableHelper.DISPOSED);
+        lifecycleDisposable.lazySet(AutoDisposableHelper.DISPOSED);
+        // Noop - we're unbound now
+      }
+    };
+    if (AutoDisposeEndConsumerHelper.setOnce(lifecycleDisposable, o, getClass())) {
+      lifecycle.subscribe(o);
       if (AutoDisposeEndConsumerHelper.setOnce(mainDisposable, d, getClass())) {
         delegate.onSubscribe(this);
```
akarnokd: Double `onSubscribe`.

autodispose/src/main/java/com/uber/autodispose/AutoDisposingSingleObserverImpl.java
```diff
@@ -41,24 +41,29 @@
...
+        callMainSubscribeIfNecessary(d);
+        mainDisposable.lazySet(AutoDisposableHelper.DISPOSED);
+        lifecycleDisposable.lazySet(AutoDisposableHelper.DISPOSED);
+        // Noop - we're unbound now
+      }
+    };
+    if (AutoDisposeEndConsumerHelper.setOnce(lifecycleDisposable, o, getClass())) {
+      lifecycle.subscribe(o);
       if (AutoDisposeEndConsumerHelper.setOnce(mainDisposable, d, getClass())) {
         delegate.onSubscribe(this);
```
akarnokd: Double `onSubscribe`.

autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java
hzsweers: As in just swap the order they're executed?

```java
mainDisposable.lazySet(AutoDisposableHelper.DISPOSED);
AutoDisposableHelper.dispose(lifecycleDisposable);
```

autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java
hzsweers: Hmm, the only events that could cause this though would be if the lifecycle terminated, in which case `mainDisposable` would either be `dispose()`'d or be lazySet to `DISPOSED` (and thus avoid a double subscription error from being thrown) right? Or do you mean there's race condition potential if the lifecycle terminates in the middle of this code execution?

hzsweers: Actually I think I see what you mean. I can call delegate.onSubscribe early rather than try to be defensive against lifecycle ending immediately (which is what I keep being worried about, as it bit me in the past). We're already subscribed at this point so it's not like calling onSubscribe on the delegate early is going to change anything

android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/LifecycleEventsObservable.java
```diff
@@ -75,14 +75,14 @@ void backfillEvents() {
...
+    observer.onSubscribe(archObserver);
     if (!isMainThread()) {
       observer.onError(
           new IllegalStateException("Lifecycles can only be bound to on the main thread!"));
       return;
     }
-    ArchLifecycleObserver archObserver =
-        new ArchLifecycleObserver(lifecycle, observer, eventsObservable);
-    observer.onSubscribe(archObserver);
     lifecycle.addObserver(archObserver);
```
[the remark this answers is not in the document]
hzsweers: a0d8556

android/autodispose-android/src/main/java/com/uber/autodispose/android/ViewAttachEventsObservable.java
```diff
@@ -46,8 +48,6 @@
       // Emit the last event, like a behavior subject
       observer.onNext(ViewLifecycleEvent.ATTACH);
     }
-    Listener listener = new Listener(view, observer);
-    observer.onSubscribe(listener);
     view.addOnAttachStateChangeListener(listener);
```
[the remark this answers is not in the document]
hzsweers: 22ba683

autodispose/src/main/java/com/uber/autodispose/AutoDispose.java
hzsweers: Good catch. Fixed in 959fe09 but planning to remove this class in the near future anyway now that `as()` is in RxJava :)

autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java
hzsweers: Done in ba8bd87

autodispose/src/main/java/com/uber/autodispose/AutoDisposingMaybeObserverImpl.java
hzsweers: part of ba8bd87

autodispose/src/main/java/com/uber/autodispose/AutoDisposingObserverImpl.java
hzsweers: part of ba8bd87

autodispose/src/main/java/com/uber/autodispose/AutoDisposingSingleObserverImpl.java
hzsweers: part of ba8bd87

autodispose/src/main/java/com/uber/autodispose/AutoDisposingMaybeObserverImpl.java
hzsweers: part of 9c3edf4

autodispose/src/main/java/com/uber/autodispose/AutoDisposingCompletableObserverImpl.java
hzsweers: Assuming so after googling a bit - part of 9c3edf4

hzsweers:
I think I've addressed the existing comments and a little extra. Thanks again for taking the time to review this, I'm learning a lot!

hzsweers:
@akarnokd let me know if there's anything else you see, otherwise I'll try to land this monday or tuesday

akarnokd:
No further comments. :+1:

hzsweers:
Awesome. Thank you so much for the review!

PR #141, opened 2017-12-03
Implement new as-based API

### LINKED ISSUE #140: Implement as()-based converter API
From the beginning, the ideal case has been for AutoDispose to be able to have a single point of entry API. `to()` unfortunately made this hard since each API accepted the same `Function` type with different generics. After contributing custom interfaces support in https://github.com/ReactiveX/RxJava/pull/5729 though, the road ahead is clear.

The plan would be for AutoDispose to remain in pre-stable until as() is promoted to stable API (possibly in RxJava 2.2).

The implementation would be straightfoward. There would be a static "entry point" method with overloads for different scopes. Under the hood they would each just return an `OmniConverter` that implements all the converter interfaces. This would be the "final" API of AutoDispose for 1.0

### Open question - What should the "entry point" method be named? 

In java 8, you won't need the generic, which means you could have APIs like this:

```java
Observable.just(1)
    .as(autoDisposable(scope))
    .subscribe();
```

For Java 7, you need the generic, and subsequently can't static import the method:

```java
Observable.just(1)
    .as(AutoDispose.<Integer>autoDisposable(scope))
    .subscribe();
```

*Note: In IntelliJ 2017.3, the [generic above will autocomplete](https://youtrack.jetbrains.com/issue/IDEA-179864)*

In the Java 7 case above, it feels a bit... redundant? Idk. It might also be nice to unify this name and the kotlin extension functions, which are currently all `autodisposeWith`

hzsweers:
This implements a new AutoDispose API (again, but final this time!) around using the new `as()` operator.

Resolves #140

Other things:
- Moves kotlin to the same name for shared semantics
- Opportunistically updated a few dependencies

Now, the API is just

```java
Observable.just(1)
    .as(autoDisposable(scope))
    .subscribe();
```

Or in Kotlin

```kotlin
Observable.just(1)
    .autoDisposable(scope)
    .subscribe();
```


hzsweers:
Note the `RestrictTo` error-prone check is disabled for now because it's totally broken (it thinks only the support library is allowed to use it)

PR #144, opened 2017-12-04
Final touchups for 0.5.0 release

hzsweers:
This is a few housecleaning changes:

- README updates
- Couple gradle fixes
- Promote `delegateObserver()` APIs to stable since we'll keep these. This ensures they're compatible with `subscribeWith` in the future, and we can keep the library flexible.


hzsweers:
Did an over the shoulder review, merging since it's not functional

Issue #146, opened 2017-12-05
Add error-prone DoNotMock annotations to elements with test helpers


hzsweers:
After fixing a bunch of cases where bad mockito usage broke tests during the last update, I think we should introduce these.

Anything with a test component should get this annotation and point to its test component.

* `ScopeProvider` -> `TestScopeProvider`
* `LifecycleScopeProvider` -> `TestLifecycleScopeProvider`

The error-prone annotations themselves should be `compileOnly` or `provided`

bangarharshit:
I can work on it. Do you also expect [positive/negative](https://github.com/google/error-prone/wiki/Writing-a-check#exercise-5-positive-cases-example-solution) test cases?

hzsweers:
No need for tests, just the annotations will do

hzsweers:
Done in #153

PR #153, opened 2017-12-31
Error prone DoNotMock annotation added to elements with test helpers

bangarharshit:
Elements with test helpers shouldn't be mocked as it can break during an update. Use Test helpers instead. Fixes - https://github.com/uber/AutoDispose/issues/146

The annotation is currently just an indication to developer as the relevant check to enforce this is not yet open sourced - https://github.com/google/error-prone/issues/572

### DIFF
--- a/autodispose/src/main/java/com/uber/autodispose/LifecycleScopeProvider.java
+++ b/autodispose/src/main/java/com/uber/autodispose/LifecycleScopeProvider.java
@@ -16,6 +16,7 @@
 
 package com.uber.autodispose;
 
+import com.google.errorprone.annotations.DoNotMock;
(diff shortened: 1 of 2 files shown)


hzsweers:
Forgot we have an internal implementation of the donotmock checker. Oh well, will take effect when someone implements one anyway.

hzsweers:
Not sure what's up with the CLA check. Will give it a bit to resolve itself

bangarharshit:
Seems to be stuck. It may need a [retry](https://github.com/cla-assistant/cla-assistant/issues/124)

hzsweers:
Thanks!

PR #155, opened 2018-01-01
Parallel flowable support

### LINKED ISSUE #142: Support ParallelFlowable
ParallelFlowable is considered a 6th distinct type in RxJava, complete with its own `as()` operator. We should probably support it.

Scope would be:
* Add to AutoDisposeConverter's supported interfaces
* Add extension functions to kotlin (no need for a deprecated `autoDisposeWith()` version
* Implementation, though I think we can just return `AutoDisposingSubscriberImpl`
  * If we do reuse it, we should add a `isParallel()` method so consumers can introspect it.

VisheshVadhera:
Closes #142

### DIFF
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDisposeConverter.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDisposeConverter.java
@@ -34,5 +34,5 @@ public interface AutoDisposeConverter<T> extends FlowableConverter<T, FlowableSu
     MaybeConverter<T, MaybeSubscribeProxy<T>>,
     SingleConverter<T, SingleSubscribeProxy<T>>,
     CompletableConverter<CompletableSubscribeProxy>,
-    ParallelFlowableConverter<T, ParallelFlowableSubscribeProxy<T>>{
+    ParallelFlowableConverter<T, ParallelFlowableSubscribeProxy<T>> {
 }
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDisposeConverter.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDisposeConverter.java
@@ -30,9 +30,9 @@ import io.reactivex.parallel.ParallelFlowableConverter;
  * @param <T> the type.
  */
 public interface AutoDisposeConverter<T> extends FlowableConverter<T, FlowableSubscribeProxy<T>>,
+    ParallelFlowableConverter<T, ParallelFlowableSubscribeProxy<T>>,
     ObservableConverter<T, ObservableSubscribeProxy<T>>,
     MaybeConverter<T, MaybeSubscribeProxy<T>>,
     SingleConverter<T, SingleSubscribeProxy<T>>,
-    CompletableConverter<CompletableSubscribeProxy>,
-    ParallelFlowableConverter<T, ParallelFlowableSubscribeProxy<T>> {
+    CompletableConverter<CompletableSubscribeProxy> {
 }
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDisposeConverter.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDisposeConverter.java
@@ -21,6 +21,7 @@ import io.reactivex.FlowableConverter;
 import io.reactivex.MaybeConverter;
 import io.reactivex.ObservableConverter;
 import io.reactivex.SingleConverter;
+import io.reactivex.parallel.ParallelFlowableConverter;
 
 /**
  * A custom converter that implements all the RxJava types converters, for use with the {@code as()}
@@ -32,5 +33,6 @@ public interface AutoDisposeConverter<T> extends FlowableConverter<T, FlowableSu
     ObservableConverter<T, ObservableSubscribeProxy<T>>,
     MaybeConverter<T, MaybeSubscribeProxy<T>>,
     SingleConverter<T, SingleSubscribeProxy<T>>,
-    CompletableConverter<CompletableSubscribeProxy> {
+    CompletableConverter<CompletableSubscribeProxy>,
+    ParallelFlowableConverter<T, ParallelFlowableSubscribeProxy<T>>{
 }
--- a/autodispose/src/main/java/com/uber/autodispose/ParallelFlowableScoper.java
+++ b/autodispose/src/main/java/com/uber/autodispose/ParallelFlowableScoper.java
@@ -6,10 +6,10 @@ import io.reactivex.Maybe;
 import io.reactivex.parallel.ParallelFlowable;
 import io.reactivex.parallel.ParallelFlowableConverter;
 
-class AutoDisposeParallelFlowableConverter<T> extends Scoper
+class ParallelFlowableScoper<T> extends Scoper
     implements ParallelFlowableConverter<T, ParallelFlowableSubscribeProxy<T>> {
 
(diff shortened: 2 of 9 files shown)


autodispose/src/main/java/com/uber/autodispose/AutoDisposeConverter.java
```diff
@@ -32,5 +33,6 @@
     ObservableConverter<T, ObservableSubscribeProxy<T>>,
     MaybeConverter<T, MaybeSubscribeProxy<T>>,
     SingleConverter<T, SingleSubscribeProxy<T>>,
-    CompletableConverter<CompletableSubscribeProxy> {
+    CompletableConverter<CompletableSubscribeProxy>,
+    ParallelFlowableConverter<T, ParallelFlowableSubscribeProxy<T>>{
```
hzsweers: nit: let's move this next to `FlowableConverter` in the declaration

autodispose/src/main/java/com/uber/autodispose/ParallelFlowableScoper.java
```diff
@@ -0,0 +1,50 @@
+package com.uber.autodispose;
+
+import org.reactivestreams.Subscriber;
+
+import io.reactivex.Maybe;
+import io.reactivex.functions.Function;
+import io.reactivex.parallel.ParallelFlowable;
+
+class ParallelFlowableScoper<T> extends Scoper
+    implements Function<ParallelFlowable<? extends T>, ParallelFlowableSubscribeProxy<T>> {
```
hzsweers: Since this is never going to be used with `to()`, let's make this implement the `ParallelFlowableConverter` directly and skip the indirection of the others. I mostly haven't migrated the others to this pattern yet to give users time to migrate, but they'll all be moved to this for 1.0. This one can start with it since it's new

autodispose/src/main/java/com/uber/autodispose/ParallelFlowableScoper.java
```diff
@@ -0,0 +1,50 @@
...
+    private final Maybe<?> scope;
+
+    AutoDisposeParallelFlowable(ParallelFlowable<T> source, Maybe<?> scope) {
+      this.source = source;
+      this.scope = scope;
+    }
+
+    @Override public void subscribe(Subscriber<? super T>[] subscribers) {
+      Subscriber<? super T>[] newSubscribers = new Subscriber[subscribers.length];
+      for (int i = 0; i < subscribers.length; i++) {
```
hzsweers: Interesting... I'm not super familiar with `ParallelFlowable`, but are there any thread safety concerns here? Is this how other `subscribe()` calls are handled in its operators?

autodispose/src/test/java/com/uber/autodispose/AutoDisposeParallelFlowableTest.java
```diff
@@ -0,0 +1,365 @@
...
+import io.reactivex.subjects.MaybeSubject;
+import io.reactivex.subscribers.TestSubscriber;
+
+import static com.google.common.truth.Truth.assertThat;
+
+public class AutoDisposeParallelFlowableTest {
+
+  private static final int DEFAULT_PARALLELISM = 2;
+
+  @Rule public RxErrorsRule rule = new RxErrorsRule();
```
hzsweers: nit: can be `final`

autodispose/src/test/java/com/uber/autodispose/AutoDisposeParallelFlowableTest.java
```diff
@@ -0,0 +1,365 @@
...
+    firstSubscriber.assertSubscribed();
+    secondSubscriber.assertSubscribed();
+
+    assertThat(source.hasSubscribers()).isTrue();
+    assertThat(lifecycle.hasObservers()).isTrue();
+
+    source.onNext(1);
+    source.onNext(2);
+    firstSubscriber.assertValue(1);
+    secondSubscriber.assertValue(2);
```
hzsweers: Can you explain this part a bit? Not quite sure I followed how this works (probably my lack of familiarity with `ParallelFlowable`).

autodispose/src/main/java/com/uber/autodispose/ParallelFlowableScoper.java
VisheshVadhera: Yep, that's what is happening in the `subscribe()` method of the operators. They iterate over the passed in array of subscribers, create a new array subscribers and pass it to the source flowable. 

One very important thing which I forgot to do was checking if the number of subscribers is equal to the parallelism of the flowable (`subscribe()'s` contract says that they should be equal.) I'll push another commit tonight.

autodispose/src/test/java/com/uber/autodispose/AutoDisposeParallelFlowableTest.java
VisheshVadhera: So the `ParallelFlowable` emits the items to the subscribers in a round robin fashion, and here we are emitting 1 & 2 to two subscribers. This means that the `firstSubscriber` will only receive 1 and the `secondSubscriber` would only receive 2 and nothing else.

autodispose/src/main/java/com/uber/autodispose/ParallelFlowableScoper.java
VisheshVadhera: Alright, I'll make the change. You want me to open an issue for this and maybe we can work on it as a part of 0.7.0 release?

hzsweers: Don't worry about it for now. I'll migrate the rest en masse for the 1.0 release when the time comes

autodispose/src/test/java/com/uber/autodispose/AutoDisposeParallelFlowableTest.java
hzsweers: gotcha :+1:

hzsweers:
Sidenote - it helps me with reviewing if you address each comment (or logical set of comments) with a single commit and then link its sha as a response to the comment

autodispose/src/main/java/com/uber/autodispose/ParallelFlowableScoper.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override public void subscribe(Subscriber<? super T>[] subscribers) {
+      if (!validate(subscribers)) {
+        return;
+      }
+
+      Subscriber<? super T>[] newSubscribers = new Subscriber[subscribers.length];
+      for (int i = 0; i < subscribers.length; i++) {
+        AutoDisposingSubscriberImpl<? super T> subscriber =
```
hzsweers: I kind of wonder if we do need a separate subscriber that manages all the subscribers here. @akarnokd penny for your thoughts?

akarnokd: Parallel rails are largely independent and act as their own Flowables. That's why many parallel operators just delegate to their sequential versions.

hzsweers: cool, in that case we can leave this as-is :+1:

hzsweers:
@VisheshVadhera do you have any other changes planned for this? (just checking)

VisheshVadhera:
@hzsweers No other changes as of now. We can land this if everything looks fine.

PR #162, opened 2018-01-21
Wiki for AutoDispose scope missing error prone check

bangarharshit:
Wiki for AutoDispose scope missing error prone check. Fixes issue - https://github.com/uber/AutoDispose/issues/152 and references commit - https://github.com/uber/AutoDispose/pull/156.

PR #167, opened 2018-02-03
Error prone check configuration flag renamed to ClassesWithScope

bangarharshit:
Based on discussions from https://github.com/uber/AutoDispose/pull/162#discussion_r165796987, renaming error prone flag to ClassesWithScope


hzsweers:
:+1: will merge once CI passes

PR #168, opened 2018-02-03
Switch to arch components common artifact for slimmer dependencies

hzsweers:
Checked with the framework folks and got confirmation that this should be ok to depend on. Now the archcomponents-test artifact no longer depends on livedata or viewmodel transitively

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -76,7 +76,7 @@ def support = [
     arch: [
         lifecycle: [
             compiler: "android.arch.lifecycle:compiler:${versions.archComponents}",
-            common: "android.arch.lifecycle:extensions:${versions.archComponents}",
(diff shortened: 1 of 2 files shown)


gradle/dependencies.gradle
```diff
@@ -76,7 +76,7 @@ def support = [
     arch: [
         lifecycle: [
             compiler: "android.arch.lifecycle:compiler:${versions.archComponents}",
-            extensions: "android.arch.lifecycle:extensions:${versions.archComponents}",
+            common: "android.arch.lifecycle:extensions:${versions.archComponents}",
```
gabrielittner: Shouldn't that be `android.arch.lifecycle:common`?

hzsweers: Derp, you're right

hzsweers: 7d16288

PR #169, opened 2018-02-03
Enable the error-prone checker for releases

hzsweers:
Now that the wiki is ready


bangarharshit:
Thanks. It looks good to me. I already disabled that GitHub account since I lost 2-FA keys.

hzsweers:
@bangarharshit you should let the org owner know that so they can remove it and use your actual account

PR #171, opened 2018-02-05
Various project build file improvements

hzsweers:
Standardizes a few declarations, uses apt-idea for better IDE support, and updates remaining project plugins to the new `plugins` DSL

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -14,6 +14,41 @@
  * limitations under the License.
  */
 
+pluginManagement {
+  resolutionStrategy {
+    eachPlugin {
+      switch (requested.id.id) {
+        case 'org.jetbrains.dokka':
+          useModule('org.jetbrains.dokka:dokka-gradle-plugin:0.9.16-eap-3')
+          break
+        case 'org.jetbrains.dokka-android':
+          useModule('org.jetbrains.dokka:dokka-android-gradle-plugin:0.9.16-eap-3')
+          break
+        case 'com.android.application':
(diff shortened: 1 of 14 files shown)


settings.gradle
```diff
@@ -14,6 +14,41 @@
  * limitations under the License.
  */
 
+pluginManagement {
+  resolutionStrategy {
+    eachPlugin {
+      switch (requested.id.id) {
+        case 'org.jetbrains.dokka':
+          useModule('org.jetbrains.dokka:dokka-gradle-plugin:0.9.16-eap-3')
```
kageiit: might be good to move these versions to deps.gradle?

Its also atypical to put this in settings.gradle. Cant it be in the root build.gradle at the top?

hzsweers: Can't apply those before settings.gradle. Want to get a baseline done first

And this has to be done in settings.gradle per the documentation - https://docs.gradle.org/current/userguide/plugins.html#sec:custom_plugin_repositories

hzsweers: If we make a buildSrc, we could do that. I'll probably just move these versions to be consolidated here though for now

kageiit: im surprised we cannot `apply from: 'deps.gradle'` in settings.gradle. Looks fine then

hzsweers: Actually let me try, I think I tried it on a different project with a more complex dependencies file than this one

hzsweers: yay it worked - 8091195

kageiit: +1

PR #174, opened 2018-02-11
Update osstrich handling in releasing

hzsweers:
This adds a helper script for running osstrich now that it's a bit more robust with kotlin projects

### DIFF
--- a/.buildscript/update_docs.sh
+++ b/.buildscript/update_docs.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+echo "Cloning osstrich..."
+mkdir tmp
+cd osstrich-tmp
+git clone git@github.com:square/osstrich.git
+cd osstrich
+echo "Packaging..."
+mvn package
+echo "Running..."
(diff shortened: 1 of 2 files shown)


.buildscript/update_docs.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+echo "Cloning osstrich..."
+mkdir tmp
+cd tmp
+git clone git@github.com:square/osstrich.git
```
kageiit: - This is going to always clone master which means builds can change in behavior. You may want to ping this
- Consider cloning this inside `~/.gradle/caches` since travis will cache that dir and avoid unnecessary fresh clones on every build everytime

hzsweers: This isn't something run on CI or otherwise automated

kageiit: oh gotcha. sry I thought i saw it on travis. nvm :)

hzsweers: :)

PR #177, opened 2018-02-16
Remove unnecessary buildconfig files from library aars

hzsweers:


android/autodispose-android/build.gradle
```diff
@@ -33,6 +33,10 @@ android {
     sourceCompatibility JavaVersion.VERSION_1_7
     targetCompatibility JavaVersion.VERSION_1_7
   }
+  // TODO replace with https://issuetracker.google.com/issues/72050365 once released.
+  libraryVariants.all {
```
kageiit: can you do it at top level build.gradle instead of duplicating please?

hzsweers: Good point, done in 5fc74b8

Note the `afterEvaluate` is necessary

PR #178, opened 2018-02-22
Add proguard rule for error prone annotations.

danh32:
Was seeing proguard build failures due to the `@DoNotMock` annotation.

This change adds an additional proguard rule to exclude the compileOnly errorprone annotations for all Android modules.

Tested by running a proguard build of Robinhood with the given rule.


danh32:
Thanks! No rush, happy to wait for the next scheduled release if nothing else big is pending.

PR #183, opened 2018-03-19
AutoDisposeAndroidPlugins (and main thread checker hook)

### LINKED ISSUE #181: New plugin hook for main thread checks
Currently, main thread checks are not configurable. In testing though, this can be a nuisance if you're trying to run in a JVM test. We've currently covered this case by just try/catching the check, but that's obviously a little wonky. A useful plugin for this would be to allow for replacing the check with your own.

From a design standpoint, the plugin would live in the `autodispose-android` artifact, and should just be modeled after [RxAndroid's plugins setup](https://github.com/ReactiveX/RxAndroid/blob/2.x/rxandroid/src/main/java/io/reactivex/android/plugins/RxAndroidPlugins.java).

The checker would be a `Callable<Boolean>` implementation. The default one can live as a static singleton instance in `AutoDisposeAndroidUtil` and initialized the [same way](https://github.com/ReactiveX/RxAndroid/blob/2.x/rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java#L27-L37) to avoid class loading (this is key so people can swap in their own impl before the `Looper` code is hit).

Method would still be `isMainThread()` in the util class, but it would defer to calling the plugin hook.

All in all, from a high level, it would look like this:

AutoDisposeAndroidUtil.java
```java
private Callable<Boolean> MAIN_THREAD_CHECK = new Callable<Boolean>() {
    @Override public Boolean call() throws Exception {
        return Looper.myLooper() == Looper.getMainLooper();
    }
}

public static boolean isMainThread() {
  return AutoDisposeAndroidPlugins.onCheckMainThread(MAIN_THREAD_CHECK);
}
```

AutoDisposePlugins.java
```java
public static boolean onCheckMainThread(Callable<Boolean> mainThreadChecker) {
  if (mainThreadChecker == null) {
    throw new NullPointerException("mainThreadChecker == null");
  }
  Callable<Boolean> c = onCheckMainThread;
  try {
    if (c == null) {
      return mainThreadChecker.call();
    } else {
      return c.call()
    }
  } catch (Exception ex) {
    throw Exceptions.propagate(ex);
  }
}

// setOnCheckMainThread(...)
// reset()
```

I don't _think_ we need the init methods their plugins impl has, but can evaluate as needed.

hzsweers:
Resolves #181 

This introduces a new plugins mechanism for Android use via `AutoDisposeAndroidPlugins`. The first plugin is a hook for main thread checks, which will allow for supplying a custom `BooleanSupplier` impl that can customize how main thread checks work. The conventional use case of this is Android JUnit tests, where the `Looper` class is not stubbed in the mock android.jar and fails explosively when touched.

One could also use this at runtime to customize checks for more fine grained main thread checks behavior.

Will update the README. See test for example usage.

### DIFF
--- a/android/autodispose-android/src/main/java/com/uber/autodispose/android/AutoDisposeAndroidPlugins.java
+++ b/android/autodispose-android/src/main/java/com/uber/autodispose/android/AutoDisposeAndroidPlugins.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright (c) 2018. Uber Technologies
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.uber.autodispose.android;
+
+import android.support.annotation.Nullable;
+import io.reactivex.exceptions.Exceptions;
+import io.reactivex.functions.BooleanSupplier;
+
+/**
+ * Utility class to inject handlers to certain standard AutoDispose Android operations.
+ */
+public final class AutoDisposeAndroidPlugins {
+
+  private AutoDisposeAndroidPlugins() { }
+
+  @Nullable private static volatile BooleanSupplier onCheckMainThread;
+
+  /**
+   * Prevents changing the plugins.
+   */
+  static volatile boolean lockdown;
+
+  /**
+   * Prevents changing the plugins from then on.
+   * <p>
+   * This allows container-like environments to prevent client messing with plugins.
+   */
+  public static void lockdown() {
+    lockdown = true;
+  }
+
+  /**
+   * Returns true if the plugins were locked down.
+   *
+   * @return true if the plugins were locked down
+   */
+  public static boolean isLockdown() {
+    return lockdown;
+  }
+
+  /**
+   * Sets the preferred main thread checker. If not {@code null}, the {@code mainThreadChecker} will
+   * be preferred in all main thread checks in {@link #onCheckMainThread(BooleanSupplier)} calls.
+   * This can be useful for JVM testing environments, where standard Android Looper APIs cannot be
+   * stubbed and thus should be overridden with a custom check.
+   * <p>
+   * This is a reset-able API, which means you can pass {@code null} as the parameter value to reset
+   * it. Alternatively, you can call {@link #reset()}.
+   *
+   * @param mainThreadChecker a {@link BooleanSupplier} to call to check if current execution is
+   * on the main thread. Should return {@code true} if it is on the main
+   * thread or {@code false} if not.
+   */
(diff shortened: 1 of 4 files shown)


hzsweers:
CC @VisheshVadhera @danh32 

Dan - want to give this a spin in y'alls use case and let me know how it works?

android/autodispose-android/src/main/java/com/uber/autodispose/android/AutoDisposeAndroidPlugins.java
```diff
@@ -0,0 +1,106 @@
...
+   * checker via {@link #setOnCheckMainThread(BooleanSupplier)} if one is present, otherwise it will
+   * use {@code defaultChecker}.
+   *
+   * @param defaultChecker the default checker to fall back to if there is no main thread checker
+   * set.
+   * @return {@code true} if called on the main thread, {@code false} if not.
+   */
+  public static boolean onCheckMainThread(BooleanSupplier defaultChecker) {
+    if (defaultChecker == null) {
+      throw new NullPointerException("defaultChecker == null");
```
tyvsmith: "defaultChecker must not be null" is a better practice for wording exceptions.

android/autodispose-android/src/main/java/com/uber/autodispose/android/AutoDisposeAndroidPlugins.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+    BooleanSupplier c = onCheckMainThread;
```
tyvsmith: more descriptive var name that c.

android/autodispose-android/src/main/java/com/uber/autodispose/android/AutoDisposeAndroidPlugins.java
```diff
@@ -0,0 +1,106 @@
...
+  /**
+   * Returns {@code true} if called on the main thread, {@code false} if not. This will prefer a set
+   * checker via {@link #setOnCheckMainThread(BooleanSupplier)} if one is present, otherwise it will
+   * use {@code defaultChecker}.
+   *
+   * @param defaultChecker the default checker to fall back to if there is no main thread checker
+   * set.
+   * @return {@code true} if called on the main thread, {@code false} if not.
+   */
+  public static boolean onCheckMainThread(BooleanSupplier defaultChecker) {
```
tyvsmith: Why is the MainThreadChecker set in a setter while the default is required in the params?

android/autodispose-android/src/main/java/com/uber/autodispose/android/AutoDisposeAndroidPlugins.java
hzsweers: we follow the rxjava convention for this here and elsewhere in autodispose, would rather discuss/address that in a followup since its a larger scope than this PR

android/autodispose-android/src/main/java/com/uber/autodispose/android/AutoDisposeAndroidPlugins.java
hzsweers: This is the RxJava/RxAndroid plugin convention and ensures that _some_ behavior is always applied even if there is no default

android/autodispose-android/src/main/java/com/uber/autodispose/android/AutoDisposeAndroidPlugins.java
hzsweers: 6f732e6

PR #186, opened 2018-03-21
Promote *SubscriberProxy extension functions from recipes to part of the Kotlin artifact

tbsandee:
These are the very useful extension functions in the recipes/samples, and I had actually built a PR porting the RxKotlin versions before I saw that these were there. I think they should be part of the autodispose-kotlin artifact to improve visibility.

I noticed that the implementations are somewhat different than their RxKotlin counterparts. I haven't been able to grok 100% why the RxKotlin versions are much more verbose. I assume some efficiency gain, but that's not clear to me. It might be worth checking over the differences to see if they are important.

https://github.com/ReactiveX/RxKotlin/blob/2.x/src/main/kotlin/io/reactivex/rxkotlin/subscribers.kt


tbsandee:
I would also be willing to contribute some test coverage, something like this perhaps (adapted from RxKotlin):
```Kotlin

  @Test
  fun observable_maybeSubscribeByExtension() {
    val first = AtomicReference<String>()
    val complete = AtomicReference<Boolean>()
    val error = AtomicReference<Throwable>()

    Observable.just("Hello")
      .autoDisposable(scopeMaybe)
      .subscribeBy(onComplete = { complete.set(true) }, onError = { error.set(it) }) {
        first.set(it)
      }

    Assert.assertSame(first.get(), "Hello")
    Assert.assertTrue(complete.get())
    Assert.assertNull(error.get())
  }
```

hzsweers:
We added these as an example for adding extensions to the proxies, but per the previous discussion around this we explicitly don't think it should be part of the library's public API (in favor of keeping the API surface area small).

PR #187, opened 2018-03-21
Fix a few nullability and other minor warnings

tbsandee:
Fix a few nullability and other minor warnings.

### DIFF
--- a/android/autodispose-android/src/main/java/com/uber/autodispose/android/ViewScopeProvider.java
+++ b/android/autodispose-android/src/main/java/com/uber/autodispose/android/ViewScopeProvider.java
@@ -16,7 +16,6 @@
 
 package com.uber.autodispose.android;
 
-import android.support.annotation.Nullable;
 import android.view.View;
 import com.uber.autodispose.LifecycleScopeProvider;
 import com.uber.autodispose.OutsideLifecycleException;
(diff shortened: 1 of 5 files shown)


android/autodispose-android/src/main/java/com/uber/autodispose/android/ViewScopeProvider.java
```diff
@@ -54,7 +55,7 @@
    * @param view the view to scope for
    * @return a {@link LifecycleScopeProvider} against this view.
    */
-  public static LifecycleScopeProvider from(View view) {
+  public static LifecycleScopeProvider<ViewLifecycleEvent> from(@Nullable View view) {
```
hzsweers: This should not be `@Nullable`. Also, let's make the return type `<?>` to keep the `ViewLifecycleEvent` an implementation detail

autodispose/src/main/java/com/uber/autodispose/ParallelFlowableScoper.java
```diff
@@ -36,7 +36,7 @@
         return;
       }
 
-      Subscriber<? super T>[] newSubscribers = new Subscriber[subscribers.length];
+      @SuppressWarnings("unchecked") Subscriber<? super T>[] newSubscribers = new Subscriber[subscribers.length];
```
hzsweers: Can this be made to not have generics issues?

tbsandee: I don't think so, because you can't create an array of a generic type T. There's always going to be some type munging behind the scenes. It's either cast here, or cast somewhere else and either way it's unchecked.

android/autodispose-android/src/main/java/com/uber/autodispose/android/ViewScopeProvider.java
tbsandee: So what triggered this was the explicit check for null -- can that be removed? I can just leave it alone with the warning too.

hzsweers: We have jsr305 package annotations to signal to static analysis that parameters are nonnull by default

We have the nullcheck as a fallback because we don't trust external users' static analysis and fail eagerly :). Just leave the warning

hzsweers:
Can you fix the checkstyle errors?

Issue #188, opened 2018-03-27
Refactor converters to be the native implementation, make old to() APIs delegate to them.


hzsweers:
So currently, every `as()`-based converter API just wraps the old `to()`-based scoper APIs. This works, but in 1.0 we're going to want to remove the old `to()` based APIs, so as an intermediary let's flip this in 0.8.0. This release'll be a non-API changing release, and totally just flip these internals as an incremental update.

These can/should be done incrementally, and should be modeled after parallelflowable's implementation. It's ok if the scopers are a bit verbose in the meantime or even duplicated functionality. The main goal here is to get the `as()` APIs fully migrated so that 1.0 is just removing the old scopers. No tests or extension artifacts (android, rxlifecycle) should be changed in this process to ensure the public API is not changing.

- [x] `ObservableConverter`/`ObservableScoper`
- [ ] `SingleConverter`/`SingleScoper`
- [ ] `MaybeConverter`/`MaybeScoper`
- [x] `CompletableConverter`/`CompletableScoper`
- [ ] `FlowableConverter`/`FlowableScoper`

hzsweers:
I'll do one of these initially to set a guideline, then try to leave the rest up for some contributors (I know there's a few people interested)

CC @atexannamedbob @VisheshVadhera @tyvsmith

hzsweers:
First one is up in #189. Once that's merged, others should feel free to claim remaining ones

VisheshVadhera:
@hzsweers #189 looks great. Would love to open a PR for the remaining ones, once the the first one gets merged.

hzsweers:
First one is down and stencil is in place. @VisheshVadhera if you could just pick one to start, want to leave the rest for others to pick up if interested :)

VisheshVadhera:
I'll pick SingleObserver.

atexannamedbob:
@hzsweers @VisheshVadhera 
I'll take on 
MaybeConverter/MaybeScoper
and
CompletableConverter/CompletableScoper

hzsweers:
I'm going to move forward with this so we can isolate the next release around it. @VisheshVadhera I'll do SingleObserver last in case you have time to come around to it, but going to do this in the next few days

VisheshVadhera:
@hzsweers Sounds good. I'll open the PR for SingleObserver tonight!

hzsweers:
This is done now

PR #189, opened 2018-03-27
Migrate ObservableScoper to delegate to as() APIs + some cleanup

hzsweers:
This is the first step of #188 and will serve as a template for the others.

- Cleans up some foundational bits to rename things in the way we want (namely `Scoper` -> `BaseAutoDisposeConverter`)
- Creates the new `AutoDisposeObservableConverter`, with pretty much identical implementation as the previous scoper but now package private and undocumented (the docs are in `AutoDispose` directly)
- Switch `AutoDispose#autoDisposable` APIs to use `AutoDisposeObservableConverter`
- Migrate `ObservableScoper` to delegate to the `autoDisposable()` method.
  - Note this makes the scoper API now a bit of overhead, which we're ok with because we want people to migrate anyway. It now goes `map()` -> `to()` -> `autoDisposable()` -> `as()`
  - The `map()` is necessary to make generics happy. I made it at least reuse a singleton instance of a function, but not keen to try to optimize this further unless someone can point out an obvious way to do this in a single step that I was missing.


hzsweers:
I took an interesting detour with this and was actually able to optimize the converters a bit. Now we inline their creation directly (we were creating one or two unnecessary allocations along the way), trimmed the code size a bit, and removed stateful intermediaries objects in an effectively synchronous flow. Squeezed more juice!

PR #190, opened 2018-03-27
Update to AGP 3.1.0 final

hzsweers:
This is a long awaited update to AGP 3.x. We had to skip 3.0 because it had gradle classpath issues and we didn't want to pollute the root build.gradle file with a bunch of plugins.

This PR does the following:
* Updates to AGP 3.1.0
* Updates dependencies to respective `api`, `implementation`, and `compileOnly` dependencies
  * Great care was taken to make these done intelligently, but let me know if there's anything that could be further narrowed or looks off.
* Enables the new [test orchestrator](https://developer.android.com/training/testing/junit-runner.html#using-android-test-orchestrator) on instrumentation tests

This also opens the door for us to add a UAST linter artifact alongside the EP checker

### DIFF
--- a/android/autodispose-android-archcomponents-kotlin/build.gradle
+++ b/android/autodispose-android-archcomponents-kotlin/build.gradle
@@ -40,9 +40,8 @@ android {
 }
 
 dependencies {
-  compile deps.kotlin.stdlib
-  compile project(':android:autodispose-android-archcomponents')
-  androidTestCompile project(':test-utils')
+  implementation project(':android:autodispose-android-archcomponents')
+  androidTestImplementation project(':test-utils')
 }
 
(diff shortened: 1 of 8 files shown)


hzsweers:
the CI bit is odd. If I run `connectedCheck` directly, the parallelism seems to fail it. If I run them individually though, they work fine. Will do some research as I'm sure this is a solved problem

hzsweers:
Blocked on Dokka - https://github.com/Kotlin/dokka/issues/282

Will need to wait for Dokka 0.9.17

hzsweers:
I'm just going to disable dokka for now, as the project seems largely unmaintained and waiting for them to update may be a fruitless endeavor

kageiit:
Agp has 3.1.2 now

PR #191, opened 2018-04-03
Migrate Completable support to fully use as()

atexannamedbob:
This is the CompletableConverter/CompletableScoper step of #188.
Converting to the new AutoDisposeCompletable and changes AutoDispose to utilize the new AutoDisposeCompletable.

### DIFF
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDisposeCompletable.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDisposeCompletable.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017. Uber Technologies
+ * Copyright (c) 2018. Uber Technologies
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


autodispose/src/main/java/com/uber/autodispose/AutoDisposeCompletable.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package com.uber.autodispose;
+
+import io.reactivex.Completable;
+import io.reactivex.CompletableObserver;
+import io.reactivex.Maybe;
+
+class AutoDisposeCompletable extends Completable {
```
hzsweers: Make this `final`

autodispose/src/main/java/com/uber/autodispose/AutoDisposeCompletable.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright (c) 2017. Uber Technologies
```
hzsweers: 2018

PR #195, opened 2018-04-27
Refactor and simplify ViewScopeProvider to just implement ScopeProvider

### LINKED ISSUE #193: ViewLifecycleEvent is exposed
```
w: /Users/pandanomic/dev/android/personal/CatchUp/app/src/debug/kotlin/io/sweers/catchup/ui/DebugViewContainer.kt: (245, 29): Type LifecycleScopeProvider<ViewLifecycleEvent!>! is inaccessible in this context due to: @RestrictTo public/*package*/ final enum class ViewLifecycleEvent : Enum<ViewLifecycleEvent!> defined in com.uber.autodispose.android
```

This is because we expose it in the generic type even though it's not public. We should fix this by reducing the scope

hzsweers:
This refactors ViewScopeProvider to just implement ScopeProvider, as it should really. It only ever considered one boundary (detached), and was a lot of overhead for it. This slims it down to just implement ScopeProvider and throw the boundary check manually.

This also reduces overhead by not eagerly subscribing to lifecycle events on the view and only creating a Maybe on demand of the detach event.

Finally - this removes the previously erroneously-exposed `ViewLifecycleEvent` enum, which was never supposed to be public.

Resolves #193


android/autodispose-android/src/main/java/com/uber/autodispose/android/internal/AutoDisposeAndroidUtil.java
```diff
@@ -34,14 +32,9 @@
...
   };
 
-  private AutoDisposeAndroidUtil() { }
+  private AutoDisposeAndroidUtil() {}
 
   public static boolean isMainThread() {
     return AutoDisposeAndroidPlugins.onCheckMainThread(MAIN_THREAD_CHECK);
   }
-
-  public static boolean isAttached(View view) {
```
jbarr21: why remove this? the API check abstraction is nice. also, why not use `ViewCompat#isAttachedToWindow(View)`?

hzsweers: We don't have a support library dependency in this artifact to keep things lean

hzsweers: Also removed just to have one less method. It was only extracted before because it was used twice

PR #196, opened 2018-04-28
Match >= for compareTo if lifecycle events are Comparable

### LINKED ISSUE #192: AutoDispose does not handle Fragment onDestroyView
Unsure if this only applies to `onDestroyView` because support fragment does some funky state manipulation, e.g. (taken from support fragment source code)

```java
void performReallyStop() {
    if (mChildFragmentManager != null) {
        mChildFragmentManager.dispatchReallyStop();
    }
    mState = ACTIVITY_CREATED;
}

void performDestroyView() {
    if (mChildFragmentManager != null) {
        mChildFragmentManager.dispatchDestroyView();
    }
    mState = CREATED;
    mCalled = false;
    onDestroyView();
    ...
}
```

Anyway, the following code causes autodispose to not crash or ever dispose of the observable.

```kotlin
override fun onDestroyView() {
    super.onDestroyView()

    Observable.interval(1, TimeUnit.SECONDS)
            .`as`(AutoDispose.autoDisposable<Long>(AndroidLifecycleScopeProvider.from(this)))
            .subscribe {
                println("Subscribed! $activity")
            }
}
```

From some initial debugging, it looks like autodispose thinks that the current lifecycle is `onStart` (which i guess makes sense because `LifecycleEventsObservable.backfillEvents` maps the state `ON_CREATE` to `ON_START`, and from looking at the fragment's code it looks like the state would be created.

hzsweers:
Resolves #192, this checks if the lifecycle even type is comparable and uses it in comparison checking. This works for enums as a convention, I can't think of any examples that don't abide by this, but also allows for overriding the behavior if necessary. The idea here is that if we target end event "B" from start event "A", any events after "B" should also be counted as stop points.

CC @jinatonic

### DIFF
--- a/autodispose/src/main/java/com/uber/autodispose/ScopeUtil.java
+++ b/autodispose/src/main/java/com/uber/autodispose/ScopeUtil.java
@@ -118,13 +118,11 @@ public final class ScopeUtil {
       Observable<E> lifecycle,
       final E endEvent) {
     Function<E, Boolean> equalityFunction;
-    if (endEvent instanceof Enum) {
-      // Match on enum ordinal in case we skip an event
-      final int targetOrdinal = ((Enum) endEvent).ordinal();
+    if (endEvent instanceof Comparable) {
       //noinspection unchecked
-      equalityFunction = (Function<E, Boolean>) new Function<Enum, Boolean>() {
-        @Override public Boolean apply(Enum e) {
-          return e.ordinal() >= targetOrdinal;
+      equalityFunction = (Function<E, Boolean>) new Function<Comparable<E>, Boolean>() {
+        @Override public Boolean apply(Comparable<E> e) {
+          return e.compareTo(endEvent) >= 0;
         }
       };
     } else {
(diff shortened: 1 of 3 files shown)


tonycosentini:
Makes sense to me - one question though, can this have adverse effects for people upgrading?

hzsweers:
Going to punt on this for the 0.8.0 release while we figure out the behavior of this a bit more. Can do a bugfix release with what we land on easily though

autodispose/src/main/java/com/uber/autodispose/ScopeUtil.java
```diff
@@ -117,12 +117,25 @@ private ScopeUtil() {
   public static <E> Maybe<LifecycleEndNotification> resolveScopeFromLifecycle(
       Observable<E> lifecycle,
       final E endEvent) {
+    Function<E, Boolean> equalityFunction;
+    if (endEvent instanceof Enum) {
+      // Match on enum ordinal in case we skip an event
+      final int targetOrdinal = ((Enum) endEvent).ordinal();
+      //noinspection unchecked
+      equalityFunction = (Function<E, Boolean>) new Function<Enum, Boolean>() {
```
tyvsmith: All java enums implement Comparable. You can use compareTo to get this behavior by default and allow consumers to provide their own if using custom enum events.

PR #219, opened 2018-07-04
Switch to errorprone-javacplugin

### LINKED ISSUE #218: Move to new error prone plugin
https://github.com/tbroyer/gradle-errorprone-javacplugin-plugin

Better compatibility with gradle moving forward, though not sure why the original wasn't just updated with these changes 🤔 

FYI @msridhar @kageiit

hzsweers:
Also opportunistically updates ErrorProne and NullAway to latest, as well as enabling nullaway on the static analysis artifact

Resolves #218

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ android:
     - platform-tools
 
 jdk:
-  - oraclejdk9
+  - oraclejdk10
 
 before_install:
   # Install SDK license so Android Gradle plugin can install deps.
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ android:
     - platform-tools
 
 jdk:
-  - oraclejdk8
+  - oraclejdk9
 
 before_install:
   # Install SDK license so Android Gradle plugin can install deps.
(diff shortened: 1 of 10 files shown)


hzsweers:
Forgot the JDK 9 requirement. Does that make sense to require for building or should we hold off?

This would definitely bork anyone in uber trying to build this since our android env mandates jdk 8 :|

.travis.yml
```diff
@@ -7,7 +7,7 @@ android:
     - platform-tools
 
 jdk:
-  - oraclejdk8
+  - oraclejdk9
```
kageiit: 😮

kageiit: you sure you want to change this?

hzsweers: Well, that's why I asked. Also it looks like we'd need to go to 10, 9 isn't even downloadable anymore

hzsweers:
Seems android can't even build with JDK 10, so going to shutter this for now

hzsweers:
Will do the EP/Nullaway updates and enabling nullaway in the EP checker separately

PR #228, opened 2018-07-23
Extract LifecycleScopeProvider to separate artifact

### LINKED ISSUE #197: Proposal: Extract LifecycleScopeProvider to separate artifact and make it implement ScopeProvider
Currently, LifecycleScopeProvider is a first class citizen in autodispose's API. I'd like to propose making it just an implementation of ScopeProvider, with preferably default handling to match the existing behavior. The goal is to simplify AutoDispose's API by reducing the surface area from 3 to 2 and allow use of LifecycleScopeProvider everywhere ScopeProvider is used.

This would be a major ish breaking change of the library, so we should decide this before 1.0

```java
public interface LifecycleScopeProvider<E> extends ScopeProvider {
  Observable<E> lifecycle();
  E peekLifecycle();
  Function<E, E> correspondingEvents();
  
  // Java 8
  default Maybe<?> requestScope() {
    // LifecycleScopeUtil would be LifecycleScopeProvider utils extracted
    return LifecycleScopeUtil.deferredResolvedLifecycle(this);
  }
}
```

By default, we could keep it java 7 compatible and have a java 8 artifact that extends via `DefaultLifecycleScopeProvider` with the `default` implementation. Alternatively, if we rewrite in kotlin, we could allow it to handle it.

Open questions for discussion:

1. Should we do this? If not, why not?
2. Should we have a default artifact?
3. If we did this in kotlin, should we still do #2 too with `@JvmDefault`?

hzsweers:
Resolves #197

This extracts `LifecycleScopeProvider` and its friends to a separate artifact and makes it extend `ScopeProvider`.

It's a bit long, but each commit is tiny and easy to digest.

Changes worth documenting:
- `LifecycleScopeProvider` is now in a separate `autodispose-lifecycle` artifact and extends `ScopeProvider`. The package name is `com.uber.autodispose.lifecycle`.
- `AutoDisposePlugins` has moved to `AutoDisposeLifecyclePlugins` since all the configuration was for lifecycle behavior. We can always bring `AutoDisposePlugins` back to the main artifact separately in the future.
- `OutsideLifecycleException` is now `OutsideScopeException`, and was _not_ moved to the new lifecycle artifact.
- `LifecycleEndNotification` is no longer public
- `ScopeUtil`'s lifecycle-related methods are now moved to `LifecycleScopes`
- `TestLifecycleScopeProvider` is moved to the new artifact
- New `autodispose-lifecycle-jdk8` artifact with a `DefaultLifecycleScopeProvider` that stubs a `default` implementation for `requestScope()`
- New `autodispose-lifecycle-ktx` artifact with extension functions for `LifecycleScopeProvider`. This `-ktx` name will be applied to the other artifacts as well (planned before 1.0).
- Deprecated APIs were not migrated from autodispose-main, as we're removing them
- `LifecycleScopeProvider` changes:
  - `correspondingEvents` now returns a more specific `Function` called `CorrespondingEventsFunction`, which can only throw an `OutsideScopeException` and simplifies the generics
  - peekLifecycle is now annotated with RxJava's `@Nullable` rather than Jetbrains


android/autodispose-android/src/test/java/com/uber/autodispose/android/AutoDisposeAndroidPluginsTest.java
```diff
@@ -20,13 +20,14 @@
...
 import org.junit.After;
+import org.junit.Before;
 import org.junit.Test;
 
 import static com.google.common.truth.Truth.assertThat;
 
 public final class AutoDisposeAndroidPluginsTest {
 
-  @After public void tearDown() {
+  @Before @After public void tearDown() {
```
tyvsmith: tearDown is probably a bad name if you're using this for before and after.

autodispose/src/main/java/com/uber/autodispose/internal/DoNotMock.java
```diff
@@ -17,7 +33,7 @@
 @Documented
 @Retention(RetentionPolicy.CLASS)
 @Target(ElementType.TYPE)
-@interface DoNotMock {
+public @interface DoNotMock {
```
tyvsmith: Can we avoid making this public, will show up on other folks classpaths and may be somewhat common.

autodispose/src/test/java/com/uber/autodispose/AutoDisposePluginsTest.java
```diff
@@ -17,44 +17,40 @@
...
 import org.junit.After;
+import org.junit.Before;
 import org.junit.Test;
 
 import static com.google.common.truth.Truth.assertThat;
 
 public final class AutoDisposePluginsTest {
 
-  @After public void tearDown() {
+  @Before @After public void tearDown() {
```
tyvsmith: Same as other tearDown comment

lifecycle/autodispose-lifecycle-ktx/src/main/kotlin/com/uber/autodispose/lifecycle/ktx/AutoDisposeLifecycle.kt
```diff
@@ -0,0 +1,82 @@
...
+import io.reactivex.annotations.CheckReturnValue
+import io.reactivex.parallel.ParallelFlowable
+
+/**
+ * Extension that proxies to [Flowable.as] + [AutoDispose.autoDisposable]
+ *
+ */
+@CheckReturnValue
+inline fun <T> Flowable<T>.autoDisposable(
+    provider: com.uber.autodispose.lifecycle.LifecycleScopeProvider<*>): FlowableSubscribeProxy<T>
```
tyvsmith: Reduce to class name

lifecycle/autodispose-lifecycle-ktx/src/test/kotlin/com/uber/autodispose/lifecycle/ktx/AutoDisposeLifecycleTest.kt
```diff
@@ -0,0 +1,318 @@
...
+  private val o = TestObserver<String>()
+  private val s = TestSubscriber<String>()
+  private val lifecycleScopeProvider = TestLifecycleScopeProvider.create()
+
+  @Test fun observable_lifecycleNotStarted() {
+    Observable.just("Hello")
+        .autoDisposable(lifecycleScopeProvider)
+        .subscribe(o)
+
+    o.assertError { it is com.uber.autodispose.lifecycle.LifecycleNotStartedException }
```
tyvsmith: shorten

lifecycle/autodispose-lifecycle-ktx/src/test/kotlin/com/uber/autodispose/lifecycle/ktx/AutoDisposeLifecycleTest.kt
```diff
@@ -0,0 +1,318 @@
...
+    lifecycleScopeProvider.start()
+    val subject = PublishSubject.create<String>()
+    subject
+        .autoDisposable(lifecycleScopeProvider)
+        .subscribe(o)
+
+    lifecycleScopeProvider.stop()
+
+    // https://github.com/ReactiveX/RxJava/issues/5178
+//    assertThat(o.isDisposed).isTrue()
```
tyvsmith: kill?

lifecycle/autodispose-lifecycle-jdk8/build.gradle
```diff
@@ -0,0 +1,46 @@
...
+
+  compileOnly deps.misc.jetbrainsAnnotations
+
+  errorprone deps.build.errorProne
+
+  testCompile project(':test-utils')
+}
+
+tasks.withType(JavaCompile) {
+  options.compilerArgs += ["-Xep:NullAway:ERROR", "-XepOpt:NullAway:AnnotatedPackages=com.uber"]
```
tyvsmith: Extract string to dedup?

lifecycle/autodispose-lifecycle/src/test/java/com/uber/autodispose/lifecycle/LifecycleScopesTest.java
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.uber.autodispose.lifecycle;
+
+public final class LifecycleScopesTest {
+
+  // TODO
```
tyvsmith: Todo for this diff?

autodispose/src/main/java/com/uber/autodispose/internal/DoNotMock.java
hzsweers: Talked offline, not really avoidable unfortunately unless we wanted to make it a separate compileOnly artifact

lifecycle/autodispose-lifecycle-ktx/src/test/kotlin/com/uber/autodispose/lifecycle/ktx/AutoDisposeLifecycleTest.kt
hzsweers: We leave these to indicate what we _would_ assert/expect there but can't due to the linked issue

lifecycle/autodispose-lifecycle-jdk8/build.gradle
hzsweers: Outside the scope of this PR

lifecycle/autodispose-lifecycle/src/test/java/com/uber/autodispose/lifecycle/LifecycleScopesTest.java
hzsweers: yep

lifecycle/autodispose-lifecycle-ktx/src/test/kotlin/com/uber/autodispose/lifecycle/ktx/AutoDisposeLifecycleTest.kt
hzsweers: Part of 7fcd5a1

lifecycle/autodispose-lifecycle-ktx/src/main/kotlin/com/uber/autodispose/lifecycle/ktx/AutoDisposeLifecycle.kt
hzsweers: Part of 7fcd5a1

autodispose/src/test/java/com/uber/autodispose/AutoDisposePluginsTest.java
hzsweers: Part of 53772ce359b7db21c15342b6edc1ff3d809e35c3

android/autodispose-android/src/test/java/com/uber/autodispose/android/AutoDisposeAndroidPluginsTest.java
hzsweers: Part of Part of 53772ce359b7db21c15342b6edc1ff3d809e35c3

hzsweers:
@kageiit @msridhar any clues to why the error-prone artifact would start failing on dependency mismatches from the changes here?

PR #230, opened 2018-07-27
Update various dependencies to latest stable

hzsweers:
Kotlin, sample libraries


gradle/dependencies.gradle
```diff
@@ -21,7 +21,7 @@ def versions = [
     dokka: '0.9.17',
     errorProne: '2.3.1',
     errorPronePlugin: '0.0.13',
-    kotlin: '1.2.21',
+    kotlin: '1.2.51',
```
0legg: Want to try 1.3-M1 in dev branch?

gradle/dependencies.gradle
```diff
@@ -21,7 +21,7 @@ def versions = [
...
     support: '27.0.2'
```
0legg: nit: doesn't match the version in sample.gradle.

gradle/dependencies.gradle
hzsweers: No - we distribute kotlin artifacts, so we can't target unstable versions

gradle/dependencies.gradle
hzsweers: Good catch - 731aa55

hzsweers:
Have considered it, but probably outside the scope of this PR. Also gradle updated in 5a71d37

gradle/dependencies.gradle
0legg: We definitely shouldn't ship it, but it can be a good idea if unstable/snapshot will target upcoming version, so we can eagerly target compatibility.

hzsweers: In general we like to keep master stable for contributors, so unlikely to check in any unstable versions :/

PR #232, opened 2018-08-01
Add MainThreadDisposable to allow disposing on the JVM

remcomokveld:
**Description**:
Due to `LifecycleEventsObservable` extending the `MainThreadDisposable` from `RxAndroid` it was not possible to test disposing of a subscription on the JVM. 

`io.reactivex.android.MainThreadObservable` also contains a check `Looper.myLooper() == Looper.getMainLooper()` which will throw an exception in JVM tests.

Using a local copy of `MainThreadObservable` which utilizes the `AutoDisposeAndroidUtil.isMainThread()` you will get more consistent "main thread check" behavior.

The unit tests that are now included only test new behavior and not the rest of the behavior of `MainThreadObservable`. First wanted to copy over tests from rxandroid, but then saw that those are robolectric, which is not a dependency in this project.

### DIFF
--- a/android/autodispose-android/src/test/java/com/uber/autodispose/android/internal/MainThreadDisposableTest.java
+++ b/android/autodispose-android/src/test/java/com/uber/autodispose/android/internal/MainThreadDisposableTest.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.uber.autodispose.android.internal;
+
+import com.uber.autodispose.android.AutoDisposeAndroidPlugins;
+import io.reactivex.functions.BooleanSupplier;
+import org.junit.Test;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import static org.junit.Assert.assertTrue;
+
+public final class MainThreadDisposableTest {
+
+  @Test public void onDisposeRunsSyncWhenMainThreadSkipped() {
+    AutoDisposeAndroidPlugins.setOnCheckMainThread(new BooleanSupplier() {
+      @Override
+      public boolean getAsBoolean() {
+        return true;
+      }
+    });
+
+    final AtomicBoolean called = new AtomicBoolean();
+
+    new MainThreadDisposable() {
+      @Override
+      protected void onDispose() {
+        called.set(true);
+      }
+    }.dispose();
+
+    assertTrue(called.get());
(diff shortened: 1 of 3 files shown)


android/autodispose-android/src/test/java/com/uber/autodispose/android/internal/MainThreadDisposableTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Test
+  public void onDisposeFailsWhenMainThreadCheckNotSet() {
+    try {
+      new MainThreadDisposable() {
+        @Override
+        protected void onDispose() { }
+      }.dispose();
+      throw new AssertionError("Expected to fail before this due to Looper not being stubbed!");
+    } catch (RuntimeException e) {
```
hzsweers: Could we make this more specific? Since this is a super specific test case, I'm kind of fine with having something specific. Even just something simple like "contains "stub"" or something simple to smoke test it

android/autodispose-android/src/test/java/com/uber/autodispose/android/internal/MainThreadDisposableTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    new MainThreadDisposable() {
+      @Override
+      protected void onDispose() {
+        called.set(true);
+      }
+    }.dispose();
+
+    assertTrue(called.get());
+    AutoDisposeAndroidPlugins.setOnCheckMainThread(null);
```
hzsweers: Nit: just put this in a before/after block in the test

```java
@Before @After public void resetPlugins() {
  AutoDisposeAndroidPlugins.reset();
}
```

android/autodispose-android/src/test/java/com/uber/autodispose/android/internal/MainThreadDisposableTest.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
hzsweers: nit: this should mention uber

android/autodispose-android/src/test/java/com/uber/autodispose/android/internal/MainThreadDisposableTest.java
```diff
@@ -0,0 +1,60 @@
...
+import io.reactivex.functions.BooleanSupplier;
+import org.junit.Test;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import static org.junit.Assert.assertTrue;
+
+public final class MainThreadDisposableTest {
+
+  @Test public void onDisposeRunsSyncWhenMainThreadSkipped() {
```
hzsweers: This title is a little unclear to me. How about `onDispose_defersToMainThreadHook`?

android/autodispose-android/src/test/java/com/uber/autodispose/android/internal/MainThreadDisposableTest.java
hzsweers: nvm, saw it's copied too

android/autodispose-android/src/test/java/com/uber/autodispose/android/internal/MainThreadDisposableTest.java
hzsweers: nvm, saw it's copied

hzsweers:
Finishing this up in #235

PR #233, opened 2018-08-01
More dependency updates and sample app cleanup

hzsweers:
- Kotlin 1.2.60
- RxLifecycle 2.2.2
- RxJava 2.2.0
- RxAndroid 2.0.2
- ConstraintLayout 1.1.2
- Use Java 8 in the sample app now that we've got desugar in AGP 3

### DIFF
--- a/sample/src/main/java/com/uber/autodispose/sample/JavaActivity.java
+++ b/sample/src/main/java/com/uber/autodispose/sample/JavaActivity.java
@@ -21,12 +21,11 @@ import android.os.Bundle;
 import android.support.annotation.Nullable;
 import android.support.v7.app.AppCompatActivity;
 import android.util.Log;
-import com.uber.autodispose.AutoDispose;
 import com.uber.autodispose.android.lifecycle.AndroidLifecycleScopeProvider;
 import io.reactivex.Observable;
-import io.reactivex.functions.Action;
-import io.reactivex.functions.Consumer;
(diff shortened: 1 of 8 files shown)


sample/src/main/java/com/uber/autodispose/sample/JavaActivity.java
```diff
@@ -89,34 +72,17 @@
...
-        .subscribe(new Consumer<Long>() {
-          @Override public void accept(Long num) {
-            Log.i(TAG, "Started in onResume(), running until in onDestroy(): " + num);
-          }
-        });
+        .doOnDispose(
+            () -> Log.i(TAG, "Disposing subscription from onResume() with untilEvent ON_DESTROY"))
+        .as(autoDisposable(AndroidLifecycleScopeProvider.from(this, Lifecycle.Event.ON_DESTROY)))
+        .subscribe(
+            num -> Log.i(TAG, "Started in onResume(), running until in onDestroy(): " + num));
```
shaishavgandhi05: Weird line wrap? Same with `doOnDispose`

hzsweers: This is kinda how google-java-format handles it. Not too weird IMO, java lambda syntax isn't the nicest IMO

shaishavgandhi05: Interesting got it!

PR #246, opened 2018-08-09
fragment & lifecycle demo

ipcjs:
**Description**:
I wrote  `JavaFragment` by mimicking `JavaActivity`.
But it doesn't work correctly, AutoDispose doesn't `dispose` at the right time.

In a Fragment, how should I make AutoDispose work with `AndroidLifecycleScopeProvider`?

**Related issue(s)**:

### DIFF
--- a/sample/src/main/AndroidManifest.xml
+++ b/sample/src/main/AndroidManifest.xml
@@ -33,6 +33,10 @@
         android:name=".JavaActivity"
         android:label="@string/app_name"
         android:theme="@style/Theme.AppCompat.Light.NoActionBar"/>
+      <activity
+        android:name=".JavaFragmentActivity"
+        android:label="@string/app_name"
+        android:theme="@style/Theme.AppCompat.Light.NoActionBar"/>
     <activity
         android:name=".KotlinActivity"
         android:label="@string/app_name_kotlin"
--- a/sample/src/main/java/com/uber/autodispose/sample/JavaFragmentActivity.java
+++ b/sample/src/main/java/com/uber/autodispose/sample/JavaFragmentActivity.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright (c) 2017. Uber Technologies
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


shaishavgandhi05:
Not sure I understand by "AutoDispose doesn't `dispose` at the right time. I tried out your sample and it seems to work fine. Once you go back from the activity, all subscriptions stop.

hzsweers:
Agreed, would like you understand what you think isn't working correctly?

shaishavgandhi05:
I think this is intended behavior. The lifecycle state in `onActivityCreated` is `CREATED`. If you check out the [LifecycleEventsObservable](https://github.com/uber/AutoDispose/blob/9313c3f30fc928f6b26beb79ac4b51cfd3cdb190/android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/LifecycleEventsObservable.java#L58) you'll see that AutoDispose maps between [Lifecycle states](https://developer.android.com/reference/androidx/lifecycle/Lifecycle.State) and the corresponding [Lifecycle.Event](https://developer.android.com/reference/androidx/lifecycle/Lifecycle.Event) so that it can reliably find the corresponding event to dispose off your subscription. 

The Lifecycle.State during `onCreate` is `INITIALIZED` which we map to `Lifecycle.Event.ON_CREATE` etc. In both `onStart` and `onActivityCreated` the Lifecycle.State is `CREATED` which would map it to an `Lifecycle.Event.ON_START`. That is my understanding and I will defer to @hzsweers if I got something wrong. 

The TL;DR of it is to use the overloaded `AndroidLifecycleScopeProvider.from(LifecycleOwner owner, Lifecycle.Event untilEvent)` to explicitly want a disposal in `onDestroy`. You can use it like so:
`AndroidLifecycleProvider.from(this, Lifecycle.Event.ON_DESTROY)`

hzsweers:
The advised solution in SDK 28+ is to use the new view lifecycle of fragments. I'll try to put together something in the sample soon-ish

hzsweers:
Closing this in favor of #260

hzsweers:
In general - yes correct behavior would be to use viewLifecycleOwner

PR #247, opened 2018-08-13
Strawman implementation of custom overloads for different RxJava types

hzsweers:
This demos overloads for accepting RxJava's other types and funneling them to the usual `Completable` factory. There's a few considerations to take here from an API design perspective:

- `Single` and `Maybe` are straight forward, basically shortcut to `ignoreElement()`.
- `Observable` and `Flowable/Publisher` get a bit weirder. There are three overloads for each:
  - Single source argument, basically a shorthand to `ignoreElements()`
  - Source + `takeFirst` boolean, basically shortcut to `take(1).ignoreElements()`
    - There are two variants of this, one that actually calls `take(1)` and has a single type argument, and the other passes through to the `takeUntil` overload but requires a second type argument for the source (more verbose).
  - Source + `takeUntil` `Predicate`, basically shortcut to `takeUntil(predicate)`
    - Requires a second type argument for the source (more verbose).
  - Should the extra parameter overloads be in the public API at all? If so, should we use `take(1)` for the `takeFirst` overload to simplify type arguments in that case?
- Should these be in the top level API at all? Should they maybe be in a `Scopes` factory utility class with better names, like `maybeScope`/`singleScope`/etc?

Not sure about this as a 1.0 goal, but could consider it. This is meant to be a discussion/proposal PR for now.

### DIFF
--- a/autodispose/src/main/java/com/uber/autodispose/AutoDispose.java
+++ b/autodispose/src/main/java/com/uber/autodispose/AutoDispose.java
@@ -22,19 +22,25 @@ import io.reactivex.CompletableSource;
 import io.reactivex.Flowable;
 import io.reactivex.Maybe;
 import io.reactivex.MaybeObserver;
+import io.reactivex.MaybeSource;
 import io.reactivex.Observable;
 import io.reactivex.ObservableConverter;
+import io.reactivex.ObservableSource;
 import io.reactivex.Observer;
 import io.reactivex.Single;
 import io.reactivex.SingleObserver;
+import io.reactivex.SingleSource;
+import io.reactivex.annotations.Nullable;
 import io.reactivex.disposables.Disposable;
 import io.reactivex.functions.Action;
 import io.reactivex.functions.BiConsumer;
 import io.reactivex.functions.Consumer;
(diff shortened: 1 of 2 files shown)


tyvsmith:
I don't think these should be in the top level API (to start). We should introduce in a Scope Util class and consider graduating some if the usage is incredibly high. 

- The single and maybe are straight forward and I don't have much to comment on those.
- For Observable and Flowable/Publisher, I don't really like the extra overloads, it feels like something the developer should have done to the stream prior to passing to Autodispose.

hzsweers:
I agree we wouldn't want all those overloads. I put them all here for demo purposes with the intention of ideally just picking one (or at least a smaller subset)

0legg:
I think we shouldn't provide overloads like this, but add some helpers (like `Transformers` we have in monorepo), so library consumers will be able to do the required changes on their own (for non-trivial cases, obviously).

hzsweers:
Like a `Scopes` class in the library? Or something even separate like RxJava2Extensions?

hzsweers:
Closing this for now

PR #248, opened 2018-08-13
Fix pom description and developer IDs

hzsweers:


hzsweers:
@tyvsmith please take a look at this as well, want to get your eyes on this before merging

PR #252, opened 2018-09-14
Add requestScope() to samples

shaishavgandhi05:
**Description**: With the move to 1.0.0-RC*, LifecycleScopeProvider extends ScopeProvider and thus needs to implement `requestScope()`. Since these are abstract classes, this wasn't caught. This fixes it.


hzsweers:
d'oh. Do we not build the sample in CI?

shaishavgandhi05:
None of the sample activities really extend those recipes. Since they're abstract, compiler was like ¯\_(ツ)_/¯

Issue #253, opened 2018-09-16
AutoDisposing ViewModel Recipe


shaishavgandhi05:
**Library version**: 1.0.0-RC2

When you're using RxJava, one common use-case with the Architecture Components `ViewModel` is that you subscribe to a stream of data in the ViewModel itself. This is beneficial say in making a network request that's in flight. Since you're subscribing in the ViewModel, the request will still get completed. You then link your ViewModel to the View with a `LiveData` or something like a `BehaviorRelay`. In such cases you'll dispose of your `Disposable` using a `CompositeDisposable` in the ViewModel and calling `dispose` in the `onCleared` of the ViewModel. 

This is a good place to bring in AutoDispose by creating a base ViewModel which implements `LifecycleScopeProvider` and automatically disposes the subscription in `onCleared`.

I have a working example which I've tested and I'd definitely like to add it to the recipes in the `sample` project. 

Another proposal (which I believe this project wouldn't go for, but hey I'll try), would be to include this recipe as part of the `arch-components` module or a separate module of it's own that can be easily imported.

hzsweers:
Let's start with a proposal PR to add it to the sample, not totally sure I followed this

hzsweers:
Done in #254

PR #254, opened 2018-09-17
Add recipe for auto disposing ViewModel

shaishavgandhi05:
**Description**: This is an initial pass at addressing #253 . The meat of the stuff is in `AutoDisposeViewModel` where we automatically dispose any streams in `onCleared`. The behavior is modeling `AndroidLifecycleProvider` and such. The only way to use it meaningfully is to extend this base class. Unfortunately, I don't see a way of using something like `ViewModelLifecyleProvider.from(viewModel)`. Let me know if that implementation looks good. 

The rest of the stuff is around constructing a sample that shows the use of the `AutoDisposeViewModel`. Again, I haven't invested that much time into it, just to see if this would be a good fit for the sample. Right now we just delay the emission of a String to simulate what a long network request would look like. The essence of the concept is that since you're subscribing in the ViewModel and then just passing on the results using a `BehaviorRelay` or `LiveData, even if config changes, your API call still continues (something that wouldn't if you had subscribed to the same observable in your Activity/Fragment since the onDestroy would cancel the call). 

If this looks good to you, we can use mockwebserver from OkHttp to throttle and simulate a network request and make it clearer. 

**Related issue(s)**: #253

### DIFF
--- a/sample/src/main/AndroidManifest.xml
+++ b/sample/src/main/AndroidManifest.xml
@@ -42,6 +42,9 @@
         android:label="@string/title_activity_arch_component"
         android:theme="@style/Theme.AppCompat.Light.NoActionBar"/>
     <activity
+        android:name=".DisposingViewModelActivity"
+        android:theme="@style/Theme.AppCompat.Light.NoActionBar"/>
+    <activity
         android:name=".HomeActivity"
         android:label="@string/app_name"
         android:theme="@style/Theme.AppCompat.Light.NoActionBar">
--- a/sample/src/main/kotlin/com/uber/autodispose/recipes/AutoDisposeViewModel.kt
+++ b/sample/src/main/kotlin/com/uber/autodispose/recipes/AutoDisposeViewModel.kt
@@ -1,0 +1,99 @@
+/*
+ * Copyright (c) 2018. Uber Technologies
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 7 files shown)


shaishavgandhi05:
So I took an initial look at doing it via mockwebserver and it looks like I'll still have to setup a bunch of stuff with Retrofit etc. since mockwebserver will simulate a network response but it needs the network to be called. 

I'm alternatively thinking of adding a request progress bar which simulates the downloading of packets of data. We can easily do that by throttling each emission via RxJava and showing the progress in a `ProgressBar` on the Activity. 

What do you think?

hzsweers:
That works too, but I don't mind having a nontrivial example as well

hzsweers:
Also re: artifact - let's hold off for now and revisit after 1.0. Having a recipe is a good start

hzsweers:
Out of curiosity, should we look at borrowing anything from here? https://gist.github.com/LouisCAD/58d3017eedb60ce00721cb32a461980f

shaishavgandhi05:
Sorry for the delay here in updating the pull request. I've been trying to workaround some mockwebserver issues with running in a non-test environment and it's not playing well. I should have something soon, either with mockwebserver or the download progress thing. 

Regarding the gist: I'm not terribly familiar with coroutines but as I understand in that gist, it lets you add cancellations to coroutines taking advantage of `Lifecycle` and adding the `GenericLifecycleObserver` to get callbacks on lifecycle changes. This seems useful when you're using coroutines in Fragments or Activities. 

As far as ViewModel goes, the general advice from Yigit and co. has been not to reference any `LifecycleOwner` in the `ViewModel`. It's susceptible to leaks since the `ViewModel` will often outlive the `LifecycleOwner` (especially in orientation changes). 

The other option is to potentially have a `ViewModel` implement a `LifecycleOwner`. It says in the docs that you can make your custom components implement `LifecycleOwner` to handle lifecycle changes. However, I don't think it'd be appropriate to do that considering that the `ViewModel` only really has two events (created and cleared/destroyed).

shaishavgandhi05:
I've updated the sample with a download progress indicator. Somehow I couldn't get mockwebserver to play well in a non test environment. It kept complaining of ports and such. Finally gave up and came up with this instead. I've added a bunch of documentation for clarity. Here's a screenshot of the screen for reference:

device-2018-10-05-153435

Issue #214, opened 2018-06-26
1.0 Breaking Changes


hzsweers:
After RxJava 2.2 is released, we'll be good to release AutoDispose 1.0. There are some breaking changes we've held off on making that we plan to make in 1.0.

- [x] Rename kotlin extension artifacts to `-ktx` to match convention
- [x] Remove all the old `Scoper` APIs 
- [x] Switch to androidx/jetpack artifacts (should line up around the same time)
- [x] Remove deprecated kotlin extension functions
- [x] Add [module docs](https://kotlinlang.org/docs/reference/kotlin-doc.html#module-and-package-documentation) for kotlin projects

jobinlawrance:
Will this include support for RxKotlin extension functions like `subscribeBy()` ?

hzsweers:
Nope! For reasons discussed in #127, we have no plans to add a `subscribeBy()` API to the public artifact.

shaishavgandhi05:
Are the checklist items open for contribution? Wouldn't mind take a crack at one of them.

hzsweers:
Yes, but not tell we have a more concrete time frame for 1.0. I think we'll try doing another release between now and then

shaishavgandhi05:
Sounds good!

shaishavgandhi05:
Switching to androix might be safe now. Google [released RC1](https://developer.android.com/topic/libraries/support-library/androidx-rn) a week ago and the public API is finalized.

hzsweers:
We might do an RC with the switched APIs, but not yet

hzsweers:
Final is out. I'm on vacation this week and at kotlinconf the next, but rough plan is to do the following simultaneous releases:

1.0.0-RC3: support 28 final, latest arch versions
1.0.0: androidx 1.0/arch 2.0

Kotlin 1.3 final as well depending on timing

shaishavgandhi05:
Haha I was just about to make a PR with the androidx changes! This plan sounds safer though. If you haven't already done the changes, I'd be happy to send the PR for after the RC3 release.

hzsweers:
This is done!

PR #256, opened 2018-09-24
Arch runtime 1.1.1, gradle 4.10.2, support 28.0.0-final, formatting

hzsweers:


hzsweers:
See: https://github.com/uber/AutoDispose/issues/214#issuecomment-423874675

PR #259, opened 2018-10-09
Migrate to AndroidX

hzsweers:


shaishavgandhi05:
Is there still a plan to do RC3 with Support library 28?

hzsweers:
Yep that's what we did!

PR #264, opened 2018-10-11
Add animal-sniffer to guard against Java 8 only APIs

### LINKED ISSUE #261: Integrate animal sniffer to prevent use of Java 8-only APIs
https://github.com/xvik/gradle-animalsniffer-plugin

shaishavgandhi05:
**Description**: Add animal-sniffer to guard against Java 8 only APIs

Resolves #261

### DIFF
--- a/autodispose-rxlifecycle/build.gradle
+++ b/autodispose-rxlifecycle/build.gradle
@@ -17,6 +17,7 @@
 plugins {
     id 'java-library'
     id 'net.ltgt.errorprone'
+    id 'ru.vyarus.animalsniffer'
 }
 
 sourceCompatibility = deps.build.javaVersion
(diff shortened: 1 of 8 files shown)


hzsweers:
We do to need to configure which java version?

shaishavgandhi05:
It's [here](https://github.com/uber/AutoDispose/pull/264/files#diff-6512f838e273b79676cac5f72456127fR43)

I've added it as 1.7 since my understanding was that animal-sniffer will check if it compiles to Java 1.7. Since we don't want any Java8-only APIs, 1.7 would work.

hzsweers:
Ah gotcha, missed that on the first read. Could we try a sample commit in this branch to confirm it fails the CI checks? Then we can revert the commit after and merge if it works as expected :+1:

shaishavgandhi05:
Pushed a commit that fails the check. Let me know if it looks good and I can revert it.

hzsweers:
Perfect! Looks good, let’s revert and we can merge

shaishavgandhi05:
Done!

PR #275, opened 2018-10-29
Deprecate Default/Kotlin LifecycleScopeProvider interfaces

hzsweers:
Let's look at just unifying this into the base class instead

TODO: benchmark kotlin's default method compilation effects compared to D8's for `default` ones, as the latter can have significant build time penalties since it requires recompilation of all implementers

If anyone in the community wants to try this out and let us know, please do!


kageiit:
> the latter can have significant build time penalties since it requires recompilation of all implementers

Not sure this is accurate. This depends on how build systems implement it. re-dexing of implementers only needs to happen if abi changes occur in D8. Not sure how the kotlin version works (I would guess the desugaring to be similar)

hzsweers:
> if abi changes occur in D8

if you change an interface, isn't that an ABI change?

shaishavgandhi05:
Want to update the sample recipes as well?

hzsweers:
I couldn't find any major build issue. At the end of the day, this would only be an issue if the interface of LifecycleScopeProvider changed, which would only be in new dependency updates and _should_ cause a rebuild anyway

PR #280, opened 2018-11-20
Update example usage of RxLifecycleInterop

MarkyC:
**Description**: The javadoc for `RxLifecycleInterop#from` was using the old way of binding the observable to AutoDispose.

**Related issue(s)**:


hzsweers:
Thanks!

Issue #203, opened 2018-04-28
Lint API artifact


hzsweers:
Alongside the error-prone tooling option, we want to add a lint option as well. This is more flexible than the error prone option and shows up in the IDE nicely.

The linter should work for both kotlin and java code, and thus also check both kotlin and java autodispose APIs. By default it should be an error.

The linter should offer a quick fix in the appropriate language API if possible (not sure how doable this is since we need to ask the user for a "scope")

It should be written in Kotlin, as it will be using UAST and this is the recommendation of the lint maintainers at Google.

This depends on #190

shaishavgandhi05:
What kind of Lint checks do we have in mind? 

Might be useful to have something like [this](https://github.com/square/sqlbrite/issues/58) which has a checklist.

hzsweers:
This is done in #282!

PR #282, opened 2018-11-22
Lint Module

shaishavgandhi05:
**Description**: Adds a lint module that checks for AutoDispose usage and raises errors when it's not being used to manage your subscriptions. 

The lint checks mainly takes advantage of the fact that when subscription is managed through AutoDispose, the `as()` operator returns an `ObservableConverter` instead of a plain old `Observable`. So we check for instances of `subscribe()` on all the five reactive types and raise an error. Let me know if that logic holds up. 

There is also a good scope for suggesting quick fixes. For example, we can find when the `subscribe()` is in an Activity/Fragment and then suggest quick fixes accordingly. I'll add that in a follow up so as to not introduce noise here. 

**Related issue(s)**: #203


static-analysis/lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Detector which checks if your stream subscriptions are handled by AutoDispose.
+ */
+class AutoDisposeDetector: Detector(), SourceCodeScanner {
+
+  companion object {
+    val ISSUE: Issue = Issue.create(
+        "AutoDisposeUsage",
+        "Subscription not managed with AutoDispose",
+        "You're subscribing to an observable but not handling it's subscription. This " +
```
shaishavgandhi05: I'm not too happy with the messages here so I'm open to suggestions!

gradle/dependencies.gradle
```diff
@@ -20,7 +20,8 @@ def versions = [
     dokka: '0.9.17',
     errorProne: '2.3.2',
     errorPronePlugin: '0.0.13',
-    kotlin: '1.3.0'
+    kotlin: '1.3.0',
+    lint: '26.2.1'
```
hzsweers: Let's use the latest - `26.3.0-beta04`. We can mark the min api in the registry via `minApi = 2`

settings.gradle
```diff
@@ -68,4 +68,5 @@ include ':lifecycle:autodispose-lifecycle-ktx'
 include 'autodispose-rxlifecycle'
 include ':test-utils'
 include ':static-analysis:autodispose-error-prone-checker'
+include ':static-analysis:lint'
```
hzsweers: let's call this `autodispose-lint` to match the naming conventions

static-analysis/lint/src/main/java/com/uber/autodispose/lint/AutoDisposeIssueRegistry.kt
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.uber.autodispose.lint
+
+import com.android.tools.lint.client.api.IssueRegistry
+import com.android.tools.lint.detector.api.Issue
+
+class AutoDisposeIssueRegistry(override val issues: List<Issue> = listOf()) : IssueRegistry() {
```
hzsweers: Let's use the autoservice registration for this

https://github.com/hzsweers/rxjava2-optionalcheckreturnvalue-checkers/blob/master/lint/src/main/kotlin/io/sweers/rxjava2optionalcheckreturnvaluechecker/lint/OptionalCheckReturnValueIssueRegistry.kt

static-analysis/lint/build.gradle
```diff
@@ -24,6 +24,14 @@ targetCompatibility = deps.build.javaVersion
...
 dependencies {
   compileOnly deps.kotlin.stdlib
-  compileOnly deps.build.lintApi
+  compileOnly deps.build.lint.api
 
+  testImplementation deps.build.lint.core
+  testImplementation deps.build.lint.tests
+}
+
+jar {
```
hzsweers: If we use the autoservice approach, we shouldn't need this I think

static-analysis/lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
hzsweers: Can we borrow the messaging from the error prone checker?

static-analysis/lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -0,0 +1,66 @@
...
+            "can result in memory leaks. You can avoid memory leaks by appending " +
+            "`.as(autoDisposable(this))` before you subscribe.",
+        Category.CORRECTNESS,
+        8,
+        Severity.ERROR,
+        Implementation(AutoDisposeDetector::class.java, EnumSet.of(Scope.JAVA_FILE, Scope.TEST_SOURCES)))
+
+    private const val OBSERVABLE = "io.reactivex.Observable"
```
hzsweers: Need to support `ParallelFlowable` as well

static-analysis/lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -0,0 +1,66 @@
...
+  override fun visitMethod(context: JavaContext, node: UCallExpression, method: PsiMethod) {
+    val evaluator = context.evaluator
+
+    if (isReactiveType(evaluator, method)) {
+      context.report(ISSUE, node, context.getLocation(node), "Subscription not managed by AutoDispose.")
+    }
+  }
+
+  private fun isReactiveType(evaluator: JavaEvaluator, method: PsiMethod): Boolean {
+    return evaluator.isMemberInClass(method, OBSERVABLE) ||
```
hzsweers: There's a bit more to be done here. Specifically, we want to _only_ apply this in scopes that have lifecycles. The error prone check does this as well if you want some prior art. Basically:

- `isReactiveType`
- "is subscribing inside a class that implements "known" lifecycle types
  - in the EP checker, we include a few "known" lifecycle types like activity/fragment/scopeprovider/etc

Then, potential icing on the cake, allow contributions of custom lifecycle types. I'm not really sure how feasible that is in lint though, so not something we need to do in this PR

static-analysis/lint/src/test/kotlin/com/uber/autodispose/lint/AutoDisposeDetectorTest.kt
```diff
@@ -0,0 +1,330 @@
...
+ */
+
+package com.uber.autodispose.lint
+
+import com.android.tools.lint.checks.infrastructure.TestFiles.java
+import com.android.tools.lint.checks.infrastructure.TestFiles.kotlin
+import com.android.tools.lint.checks.infrastructure.TestLintTask.lint
+import org.junit.Test
+
+class AutoDisposeDetectorTest {
```
hzsweers: Crazy idea - could we reuse the same test files for this and the error prone check? Can make a third module just to hold those resources for tests

shaishavgandhi05:
I'm guessing we do rebase without merge commits? I think I messed up the commit history trying to do that.

hzsweers:
yeah always rebase and force push

static-analysis/lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
shaishavgandhi05: Should it only be applied for scopes that have lifecycles? For example, if you're using a Presenter/ViewModel/Controller etc. those are all still valid cases where lifecycle should be handled.

static-analysis/lint/src/main/java/com/uber/autodispose/lint/AutoDisposeIssueRegistry.kt
shaishavgandhi05: Nice! That lint registry always felt like the most awkward part.

static-analysis/lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
hzsweers: Yep, only places that have scopes. Those cases would either need to implement ScopeProvider or (if we can accept custom types) specified separately

static-analysis/lint/src/test/kotlin/com/uber/autodispose/lint/AutoDisposeDetectorTest.kt
shaishavgandhi05: Oh nice thought! Let me try if that works!

shaishavgandhi05:
> What does integration look like outside of android projects? It can be done via putting the checks on the annotationprocessor classpath but my memory's hazy, and we'll want to make sure we cover this case too.
Have you tried this in a consuming project? Did you see the lint get picked up? Screenshot if so? :)

I haven't tried it yet for a non android project. I can try and report back. 
As for consuming it, I've tried it in our sample project and it works on the command line. I'm getting different accounts for IDE integration with the [latest](https://groups.google.com/forum/#!topic/lint-dev/HbTSx-dZ2JA) being that custom lint checks are supported from 3.3 beta4 onwards. I'm unclear whether that happens automatically or there is some setup required from the user's side. 

P.S - The CI failed coz the lint check worked 😛

hzsweers:
I think the implication there was just that it was broken during 3.3 alpha/betas consuming them (I commented asking for confirmation), as custom lint checks like Timber's have showed up for me for a long time in the past. Assuming it is a 3.3 regression, I'm happy to also basically acknowledge the bugs in 3.3. You should be able to try the workarounds listed though right?

shaishavgandhi05:
Ooo super interesting. After copying the lint jar file into the `/.android/lint` directory, the check still wouldn't show up on the IDE. After digging around a little, I found [this](https://groups.google.com/forum/#!searchin/lint-dev/ide%7Csort:date/lint-dev/ULQMzW1ZlP0/mjVUmu0vDgAJ) thread which mentions that even if a JAR is in that module, it will run "on-the-fly" only if it's considered light enough by Android Studio. This is determined by the `Scope` that is provided in the `Implementation`. If I remove the `Scope.TEST_SOURCES` from the current `EnumSet.of(Scope.JAVA_FILE, Scope.TEST_SOURCES)`, it works in the IDE. 

According to me, that seems fair, since we don't want to necessarily run it for tests. What do you think?

shaishavgandhi05:
Before
screen shot 2018-11-22 at 9 37 21 pm

After
screen shot 2018-11-22 at 9 38 13 pm

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -51,16 +56,27 @@ class AutoDisposeDetector: Detector(), SourceCodeScanner {
...
+    if (isReactiveType(evaluator, method) && isInScope(evaluator, node.getContainingUClass())) {
+      context.report(ISSUE, node, context.getLocation(node), "Always apply an AutoDispose " +
+          "scope before subscribing within defined scoped elements.")
     }
   }
 
+  private fun isInScope(evaluator: JavaEvaluator, psiClass: PsiClass?): Boolean {
+    psiClass?.let {
+      return evaluator.inheritsFrom(psiClass, "androidx.lifecycle.LifecycleOwner", false) ||
+          evaluator.inheritsFrom(psiClass, "com.uber.autodispose.ScopeProvider", false) ||
```
hzsweers: let's expand this to also cover `Activity` and `Fragment` (old and new). Specifically let's try to use the same default coverage as the error prone check, and stick this in something a bit more DRY'd up (`defaultScopes.any { evaluator.inheritsFrom(psiClass, it) }`)

hzsweers:
> If I remove the Scope.TEST_SOURCES from the current EnumSet.of(Scope.JAVA_FILE, Scope.TEST_SOURCES), it works in the IDE.

I think that's fair, though I've raised a question on that linked thread asking if this behavior needs to be limited to non-test sources in general, as that seems potentially harsh

Sent this to you over chat, but for posterity:

```
also check out https://groups.google.com/d/msg/lint-dev/JxthreZ_Ny8/BuDHkcHKAAAJ
re: configuration of lints
doesn't have to be gradle properties, we could make a lint.properties file instead and say that's supported
be sure to read Tor's comment below it
```

hzsweers:
And for configuration, @vanniktech reminded me that there's a separate gradle plugin (`com.android.lint`) for wiring lint checks in plain old java/kotlin projects that adds a `lintChecks` configuration, so that _should_ cover it. I think that's only for consumers though, whereas the configuration in the android plugin is for specifying which ones should be packaged in? Will ask on lint-dev (or maybe @vanniktech knows?)

Actually - @vanniktech, want to do a review of this too?

hzsweers:
From friend of the project @remcomokveld - you could try something like this to get properties into tests

```kotlin
val properties = projectProperties()
properties.property(key, value)
properties.to("gradle.properties")
lint().files(
    properties,
    kotlin(...)
)
```

autodispose/build.gradle
```diff
@@ -26,13 +27,19 @@ test {
...
+}
+
 dependencies {
   annotationProcessor deps.build.nullAway
   testAnnotationProcessor deps.build.nullAway
 
   api deps.rx.java
 
   errorprone deps.build.errorProne
+  lintChecks project(':static-analysis:autodispose-lint')
```
vanniktech: I haven't gotten this to **actually** report anything though - https://issuetracker.google.com/issues/112526243

sample/build.gradle
```diff
@@ -34,6 +34,10 @@ android {
   sourceSets {
     main.java.srcDirs += 'src/main/kotlin'
   }
+
+  lintOptions {
+    abortOnError true
```
vanniktech: This is the default one. So no need to configure it

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Detector which checks if your stream subscriptions are handled by AutoDispose.
+ */
+class AutoDisposeDetector: Detector(), SourceCodeScanner {
+
+  companion object {
+    val ISSUE: Issue = Issue.create(
+        "AutoDisposeUsage",
+        "Always apply an AutoDispose scope before subscribing within defined scoped elements.",
```
vanniktech: Typically this should be shorter since it shows up in the Android Studio IDE

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -0,0 +1,82 @@
...
+    psiClass?.let {
+      return evaluator.inheritsFrom(psiClass, "androidx.lifecycle.LifecycleOwner", false) ||
+          evaluator.inheritsFrom(psiClass, "com.uber.autodispose.ScopeProvider", false) ||
+          evaluator.inheritsFrom(psiClass, "com.uber.autodispose.lifecycle.LifecycleScopeProvider", false)
+    }
+    return false
+  }
+
+  private fun isReactiveType(evaluator: JavaEvaluator, method: PsiMethod): Boolean {
+    return evaluator.isMemberInClass(method, OBSERVABLE) ||
```
vanniktech: Could be using `listOf(<pass_all_the_types>).any { evaluator.isMemberInCLass(method, it)`

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeIssueRegistry.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import com.android.tools.lint.client.api.IssueRegistry
+import com.android.tools.lint.detector.api.CURRENT_API
+import com.android.tools.lint.detector.api.Issue
+import com.google.auto.service.AutoService
+
+@AutoService(IssueRegistry::class)
+class AutoDisposeIssueRegistry(override val issues: List<Issue> = listOf(AutoDisposeDetector.ISSUE)) : IssueRegistry() {
+  override val api: Int = CURRENT_API
+  override val minApi = 2
```
vanniktech: Does it really need `minApi`? I've also noticed this in the checks from Zac

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeIssueRegistry.kt
```diff
@@ -0,0 +1,28 @@
...
+
+package com.uber.autodispose.lint
+
+import com.android.tools.lint.client.api.IssueRegistry
+import com.android.tools.lint.detector.api.CURRENT_API
+import com.android.tools.lint.detector.api.Issue
+import com.google.auto.service.AutoService
+
+@AutoService(IssueRegistry::class)
+class AutoDisposeIssueRegistry(override val issues: List<Issue> = listOf(AutoDisposeDetector.ISSUE)) : IssueRegistry() {
```
vanniktech: might be worth putting the issues into it's own property and not into the constructor as possibly you'll get more lint checks and the line is already long

static-analysis/autodispose-lint/src/test/kotlin/com/uber/autodispose/lint/AutoDisposeDetectorTest.kt
```diff
@@ -0,0 +1,375 @@
...
+package com.uber.autodispose.lint
+
+import com.android.tools.lint.checks.infrastructure.TestFiles.java
+import com.android.tools.lint.checks.infrastructure.TestFiles.kotlin
+import com.android.tools.lint.checks.infrastructure.TestLintTask.lint
+import org.junit.Test
+
+class AutoDisposeDetectorTest {
+
+  companion object {
```
vanniktech: nit: no need for companion object

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeIssueRegistry.kt
shaishavgandhi05: With 26.3.0-beta04, the CURRENT_API actually points to AGP 3.3.0-alpha12+. Since this project still uses 3.2.1, we need `minApi = 2` so that it runs for `3.2.0-alpha7+`. We can remove it after we move to `3.3.0` stable.

sample/build.gradle
shaishavgandhi05: Fixed in 26576f1577197783703616ad4123e5be147b7ad3

shaishavgandhi05:
So I've added support for custom scopes in lint configuration along with some tests. It works fine, however, the property file access is very restricted. From the `project.propertyFiles` code:
```java

@NonNull
    public List<File> getPropertyFiles() {
        List<File> propertyFiles = new ArrayList<>(2);
        File local = new File(dir, FN_LOCAL_PROPERTIES);
        if (local.isFile()) {
            propertyFiles.add(local);
        }
        File gradle = new File(dir, FN_GRADLE_PROPERTIES);
        if (gradle.isFile()) {
            propertyFiles.add(gradle);
        }
        File wrapper = new File(dir, FD_GRADLE_WRAPPER + separator + FN_GRADLE_WRAPPER_PROPERTIES);
        if (wrapper.isFile()) {
            propertyFiles.add(wrapper);
        }
        return propertyFiles;
    }
```

As you can see, it only filters to give `local.properties`/`gradle.properties`/`gradle-wrapper.properties` so we can't read a `lint.properties` as such. 

This will probably be redundant in a follow-up PR where we read from `@HasScope` annotation and so it's not a deal-breaker. I would, however, prefer it to be `gradle.properties` since `local.properties` is often in the `.gitignore` and so wouldn't actually make it to a VCS.

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeIssueRegistry.kt
shaishavgandhi05: Resolved with 49b8ccd9455b36cc785ea5aee0f29f3e3185a41d

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
shaishavgandhi05: Resolved with 49b8ccd9455b36cc785ea5aee0f29f3e3185a41d

static-analysis/autodispose-lint/src/test/kotlin/com/uber/autodispose/lint/AutoDisposeDetectorTest.kt
shaishavgandhi05: I've kept it around so in future we could possibly reuse those stubs.

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
shaishavgandhi05: Hmmm it doesn't seem too long in the screenshot I attached in the thread above. Basically copied the message from the error-prone check. @hzsweers Any preference?

hzsweers:
Interesting, I suppose that'll have to do for now. Will do another pass on the PR as well

autodispose/build.gradle
hzsweers: I thought this was just for projects consuming the lints? Not for packaging

autodispose/src/main/java/com/uber/autodispose/TestScopeProvider.java
```diff
@@ -48,6 +48,7 @@ public static TestScopeProvider create(Completable delegate) {
 
   private final CompletableSubject innerScope = CompletableSubject.create();
 
+  @SuppressWarnings("AutoDisposeUsage")
```
hzsweers: Meta.

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -0,0 +1,101 @@
...
+import com.android.tools.lint.client.api.JavaEvaluator
+import com.android.tools.lint.detector.api.*
+import com.intellij.psi.PsiClass
+import com.intellij.psi.PsiMethod
+import org.jetbrains.uast.UCallExpression
+import org.jetbrains.uast.getContainingClass
+import org.jetbrains.uast.getContainingUClass
+import java.io.StringReader
+import java.lang.IllegalStateException
+import java.util.*
```
hzsweers: nit: no wildcards

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
hzsweers: Let's keep it as is, maybe with a linebreak after the first sentence? The IDE does the thing where it says you can read more right?

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    private const val OBSERVABLE = "io.reactivex.Observable"
+    private const val FLOWABLE = "io.reactivex.Flowable"
+    private const val PARALLEL_FLOWABLE = "io.reactivex.parallel.ParallelFlowable"
+    private const val SINGLE = "io.reactivex.Single"
+    private const val MAYBE = "io.reactivex.Maybe"
+    private const val COMPLETABLE = "io.reactivex.Completable"
+  }
+
+  private val defaultScopes = mutableSetOf("androidx.lifecycle.LifecycleOwner",
```
hzsweers: nit: move these to the `companion object` since they're constant

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -0,0 +1,101 @@
...
+      "com.uber.autodispose.ScopeProvider",
+      "com.uber.autodispose.lifecycle.LifecycleScopeProvider",
+      "android.app.Activity",
+      "android.app.Fragment")
+
+  override fun beforeCheckRootProject(context: Context) {
+    val props = Properties()
+    context.project.propertyFiles.find { it.name == "local.properties" }?.apply {
```
hzsweers: do we have to hardcode to `local.properties`? Those aren't usually checked in, if anything we'd want `gradle.properties` for the time being

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -0,0 +1,101 @@
...
+      val content = StringReader(context.client.readFile(this).toString())
+      props.load(content)
```
hzsweers: do you need to wrap this in `use` for resource cleanup? Or better yet, can we skip the string reader and do something that purely just streams it?

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -0,0 +1,101 @@
...
+      val scopes = props.getProperty(CUSTOM_SCOPE_KEY).split(",").map { it.trim() }
+      defaultScopes.addAll(scopes)
+    }
+  }
+
+  override fun getApplicableMethodNames(): List<String> = listOf("subscribe")
```
hzsweers: subscribeWith should be covered too right? (though note it doesn't exist for `ParallelFlowable`)

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -0,0 +1,101 @@
...
+
+  override fun visitMethod(context: JavaContext, node: UCallExpression, method: PsiMethod) {
+    val evaluator = context.evaluator
+
+    if (isReactiveType(evaluator, method) && isInScope(evaluator, node.getContainingUClass())) {
+      context.report(ISSUE, node, context.getLocation(node), "Always apply an AutoDispose " +
```
hzsweers: Can we reuse the messages from what's described in the implementation above?

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -0,0 +1,101 @@
...
+          "scope before subscribing within defined scoped elements.")
+    }
+  }
+
+  private fun isInScope(evaluator: JavaEvaluator, psiClass: PsiClass?): Boolean {
```
hzsweers: Mind putting a short doc on this function?

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -0,0 +1,101 @@
...
+    psiClass?.let { callingClass ->
+      return defaultScopes.any {
+        evaluator.inheritsFrom(callingClass, it, false)
+      }
+    }
+    return false
+  }
+
+  private fun isReactiveType(evaluator: JavaEvaluator, method: PsiMethod): Boolean {
+    return evaluator.isMemberInClass(method, OBSERVABLE) ||
```
hzsweers: Maybe we can put all the types in a set and just do a `rxTypes.any { evaluator.isMemeberInClass(method, it) }`?

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeIssueRegistry.kt
hzsweers: yeah I was doing this just going off of a thread I saw in lint-dev to basically support a lower lint version while still targeting newer APIs

static-analysis/autodispose-lint/src/test/kotlin/com/uber/autodispose/lint/LintFilesExtension.kt
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.uber.autodispose.lint
+
+import com.android.tools.lint.checks.infrastructure.TestFiles.bytes
+
+fun Any.rxJava2() = bytes("libs/rxjava-2.1.7.jar", javaClass.getResourceAsStream("/rxjava-2.1.7.jar").readBytes())
```
hzsweers: This is clever, but I also wonder if there isn't maybe a better way for this, like yanking it from a gradle dependency. Not important to figure out for this PR though

hzsweers:
Also - mind filling a FR in lint-dev for custom properties files support?

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
shaishavgandhi05: Yeah I was waiting on your clearance to move to `gradle.properties`. Should we allow any `.properties` file or constrain it to `gradle.properties`?

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
shaishavgandhi05: IntelliJ 🙄

autodispose/build.gradle
shaishavgandhi05: It's also for packaging (prior art in [Butterknife](https://github.com/JakeWharton/butterknife/blob/master/butterknife-runtime/build.gradle#L40)) in a library.

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
shaishavgandhi05: It's only one sentence though. The way it works is `Always apply an AutoDispose scope before subscribing within defined scoped elements.` comes as the primary hint. When you click "Read More", you're given the detailed message in line 41. `You're subscribing to an observable but not handling it's subscription. This can result in memory leaks. You can avoid memory leaks by appending .as(autoDisposable(this)) before you subscribe.`

shaishavgandhi05:
[FR filed for custom properties files](https://groups.google.com/forum/#!topic/lint-dev/AFPDRnsKCL0)

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
shaishavgandhi05: Do we really need the stream? From Tor's [comment](https://groups.google.com/forum/#!msg/lint-dev/JxthreZ_Ny8/BuDHkcHKAAAJ) in the thread, it seemed that going with a StringReader was fine. What's the benefit of instead using something like `FileInputStream` since our contents are still being loaded by the `readFile` convenience method?

hzsweers: Ah good catch, can leave as-is then 👍

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -76,11 +81,24 @@ class AutoDisposeDetector: Detector(), SourceCodeScanner {
...
+  /**
+   * Checks if the calling method is in "scope" that can be handled by AutoDispose.
+   *
+   * If your subscribe/subscribeWith method is called in a scope
+   * that is recognized by AutoDispose, this returns true. This indicates that
+   * you're subscribing in a scope and therefore, you must handle the subscription.
+   * Default scopes include Android activities, fragments and custom classes that
+   * implement ScopeProvider.
+   *
+   * @see defaultScopes
```
hzsweers: nit: this should go last

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -76,11 +81,24 @@ class AutoDisposeDetector: Detector(), SourceCodeScanner {
...
-      context.report(ISSUE, node, context.getLocation(node), "Always apply an AutoDispose " +
-          "scope before subscribing within defined scoped elements.")
+      context.report(ISSUE, node, context.getLocation(node), LINT_DESCRIPTION)
     }
   }
 
+  /**
+   * Checks if the calling method is in "scope" that can be handled by AutoDispose.
+   *
+   * If your subscribe/subscribeWith method is called in a scope
```
hzsweers: nit: backtick these since they're methods

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -63,11 +63,13 @@ class AutoDisposeDetector: Detector(), SourceCodeScanner {
 
     private val reactiveTypes = mutableSetOf(OBSERVABLE, FLOWABLE, PARALLEL_FLOWABLE, SINGLE, MAYBE,
         COMPLETABLE)
+
+    const val PROPERTY_FILE = "gradle.properties"
```
hzsweers: nit: private

static-analysis/autodispose-lint/src/test/kotlin/com/uber/autodispose/lint/AutoDisposeDetectorTest.kt
```diff
@@ -112,6 +114,74 @@ class AutoDisposeDetectorTest {
...
+    lint()
+        .files(rxJava2(), LIFECYCLE_OWNER, FRAGMENT, java("""
+          package foo;
+          import io.reactivex.Observable;
+          import io.reactivex.observers.DisposableObserver;
+          import androidx.fragment.app.Fragment;
+
+          class ExampleClass extends Fragment {
+            void names() {
+              Observable obs = Observable.just(1, 2, 3, 4);
```
hzsweers: nit: use proper generics

static-analysis/autodispose-lint/src/test/kotlin/com/uber/autodispose/lint/AutoDisposeDetectorTest.kt
```diff
@@ -112,6 +114,74 @@ class AutoDisposeDetectorTest {
...
+          package foo;
+          import io.reactivex.Observable;
+          import io.reactivex.observers.DisposableObserver;
+          import androidx.fragment.app.Fragment;
+          import com.uber.autodispose.ScopeProvider;
+
+          class ExampleClass extends Fragment {
+            ScopeProvider scopeProvider;
+            void names() {
+              Observable obs = Observable.just(1, 2, 3, 4);
```
hzsweers: nit: use proper generics

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -72,8 +72,12 @@ class AutoDisposeDetector: Detector(), SourceCodeScanner {
     context.project.propertyFiles.find { it.name == PROPERTY_FILE }?.apply {
       val content = StringReader(context.client.readFile(this).toString())
       props.load(content)
-      val scopes = props.getProperty(CUSTOM_SCOPE_KEY).split(",").map { it.trim() }
-      defaultScopes.addAll(scopes)
+      props.getProperty(CUSTOM_SCOPE_KEY)?.let { scopeProperty ->
```
hzsweers: can we add a test for this?

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
shaishavgandhi05: Will change to `internal`. Reusing this in tests.

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
shaishavgandhi05: 3ed7031e5da8c95f6723e3d2707d7e30291f7400

static-analysis/autodispose-lint/src/test/kotlin/com/uber/autodispose/lint/AutoDisposeDetectorTest.kt
shaishavgandhi05: 06d7324829c34b6fcbb0ceb34b62597f45d2a7e8

static-analysis/autodispose-lint/src/test/kotlin/com/uber/autodispose/lint/AutoDisposeDetectorTest.kt
shaishavgandhi05: 06d7324829c34b6fcbb0ceb34b62597f45d2a7e8

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
shaishavgandhi05: 06d7324829c34b6fcbb0ceb34b62597f45d2a7e8

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
shaishavgandhi05: 06d7324829c34b6fcbb0ceb34b62597f45d2a7e8

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
shaishavgandhi05: 06d7324829c34b6fcbb0ceb34b62597f45d2a7e8

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -55,19 +55,29 @@ class AutoDisposeDetector: Detector(), SourceCodeScanner {
...
+    private val defaultScopes = listOf("androidx.lifecycle.LifecycleOwner",
         "com.uber.autodispose.ScopeProvider",
         "com.uber.autodispose.lifecycle.LifecycleScopeProvider",
         "android.app.Activity",
         "android.app.Fragment")
 
+    // The scopes that are applicable for the lint check.
+    // This includes the defaultScopes as well as any custom scopes
+    // defined by the consumer.
+    private val appliedScopes = mutableSetOf<String>()
```
hzsweers: This is still static state, let's move it down to be an instance property

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -56,26 +56,26 @@ class AutoDisposeDetector: Detector(), SourceCodeScanner {
...
+    private val REACTIVE_TYPES = mutableSetOf(OBSERVABLE, FLOWABLE, PARALLEL_FLOWABLE, SINGLE, MAYBE,
         COMPLETABLE)
 
     internal const val PROPERTY_FILE = "gradle.properties"
   }
 
+  // The scopes that are applicable for the lint check.
+  // This includes the DEFAULT_SCOPES as well as any custom scopes
+  // defined by the consumer.
+  private val appliedScopes = mutableSetOf<String>()
```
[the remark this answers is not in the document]
shaishavgandhi05: If we do that, we can make it an immutable set.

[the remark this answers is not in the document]
hzsweers: yep!

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -56,26 +56,26 @@ class AutoDisposeDetector: Detector(), SourceCodeScanner {
...
         "android.app.Activity",
         "android.app.Fragment")
 
-    // The scopes that are applicable for the lint check.
-    // This includes the defaultScopes as well as any custom scopes
-    // defined by the consumer.
-    private val appliedScopes = mutableSetOf<String>()
-
-    private val reactiveTypes = mutableSetOf(OBSERVABLE, FLOWABLE, PARALLEL_FLOWABLE, SINGLE, MAYBE,
+    private val REACTIVE_TYPES = mutableSetOf(OBSERVABLE, FLOWABLE, PARALLEL_FLOWABLE, SINGLE, MAYBE,
```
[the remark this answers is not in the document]
shaishavgandhi05: c355ed3843a034d46b40252718a360dd069936f6

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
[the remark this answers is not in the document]
shaishavgandhi05: c355ed3843a034d46b40252718a360dd069936f6

Issue #287, opened 2018-11-25
Error Prone checker is missing service file


hzsweers:
Because autoservice is not applied in the `annotationProcessor` configuration, meaning it's never worked!

PR #289, opened 2018-11-25
Apply autoservice processing to error prone checker

shaishavgandhi05:
**Description**: This adds the missing service file to the error-prone JAR.

**Related issue(s)**: #287 

What do you think about integrating this into the sample project, so that we can easily test newer/existing checks?


shaishavgandhi05:
Wrt adding to sample, I see #218 now and seems like that's on hold

hzsweers:
Talked offline, this is being taken care of in a larger change

PR #303, opened 2018-12-03
Rename autodispose-error-prone-checker to just autodispose-error-prone

hzsweers:
To match the semantics of autodispose-lint


shaishavgandhi05:
Also update in the Wiki after merge re artifact name

PR #307, opened 2018-12-05
Lint Lenient Mode for captured Disposable

### LINKED ISSUE #295: Support lenient mode in lint
See the supported mode details in #291

We can borrow from the CheckReturn lint check the same way the EP check borrow's EP's CheckReturnValue logic, example here: https://github.com/hzsweers/rxjava2-optionalcheckreturnvalue-checkers/blob/bb2efe03e6903dd83943aee1e74a4074aec4a409/lint/src/main/kotlin/io/sweers/rxjava2optionalcheckreturnvaluechecker/lint/OptionalCheckReturnValueDetector.kt

Note that the EP check checks a wider range of stuff, but I'm fine with just limiting the check to what the existing check return lint checks (aka I don't think it checks for things in tests, mockito, etc)

shaishavgandhi05:
**Description**: This checks if the returned `Disposable` is handled by the user and doesn't raise errors in that case. We assume that you know what you're doing with it. 

Resolves #295

### DIFF
--- a/static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
+++ b/static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
@@ -17,10 +17,26 @@
 package com.uber.autodispose.lint
 
 import com.android.tools.lint.client.api.JavaEvaluator
-import com.android.tools.lint.detector.api.*
+import com.android.tools.lint.detector.api.Category
+import com.android.tools.lint.detector.api.Detector
+import com.android.tools.lint.detector.api.Issue
+import com.android.tools.lint.detector.api.SourceCodeScanner
+import com.android.tools.lint.detector.api.Severity
+import com.android.tools.lint.detector.api.Implementation
+import com.android.tools.lint.detector.api.Scope
+import com.android.tools.lint.detector.api.JavaContext
+import com.android.tools.lint.detector.api.Context
 import com.intellij.psi.PsiClass
 import com.intellij.psi.PsiMethod
 import org.jetbrains.uast.UCallExpression
+import org.jetbrains.uast.UElement
+import org.jetbrains.uast.getParentOfType
+import org.jetbrains.uast.UExpression
+import org.jetbrains.uast.UQualifiedReferenceExpression
+import org.jetbrains.uast.ULambdaExpression
+import org.jetbrains.uast.UMethod
+import org.jetbrains.uast.UClassInitializer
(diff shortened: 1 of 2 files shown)


shaishavgandhi05:
It does. I can add a test case.

static-analysis/autodispose-lint/src/test/kotlin/com/uber/autodispose/lint/AutoDisposeDetectorTest.kt
```diff
@@ -621,4 +621,36 @@ class AutoDisposeDetectorTest {
         .run()
         .expectClean()
   }
+
+  @Test fun subscribeWithCapturedDisposable() {
```
hzsweers: We'll need failure cases here too. Can you try to reuse some of the tests we had for the error prone check?

shaishavgandhi05: We already have a failing `subscribeWith` here https://github.com/uber/AutoDispose/blob/986b4b206b75926bcf2ba514f5bccb0ef9ec504d/static-analysis/autodispose-lint/src/test/kotlin/com/uber/autodispose/lint/AutoDisposeDetectorTest.kt#L117

hzsweers: I meant capturing a returned type that doesn't implement `Disposable`

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -126,9 +128,13 @@ class AutoDisposeDetector: Detector(), SourceCodeScanner {
 
     if (isReactiveType(evaluator, method)
         && isInScope(evaluator, node.getContainingUClass())
-        && isExpressionValueUnused(node)
     ) {
-      context.report(ISSUE, node, context.getLocation(node), LINT_DESCRIPTION)
+      val isUnusedReturnValue = isExpressionValueUnused(node)
+      if (isUnusedReturnValue || !isCapturedTypeAllowed(node.returnType, method)) {
```
hzsweers: Can you add a test for method references? Both in java and kotlin

shaishavgandhi05: 443573d7e5f85bcb60869d9aeec69f4c2c8eb5d4

Issue #154, opened 2017-12-31
Code formatter tool to adhere to checkstyle


bangarharshit:
[uber/NullAway](https://github.com/uber/NullAway/blob/master/config/hooks/pre-commit) uses a pre-commit hook to format code to google Java style. Can be pretty useful to contributors as they don't have to worry about styling.

hzsweers:
does it require the user to manually install it? I'd been leaning toward using Spotless since that also covers kotlin

bangarharshit:
It doesn't require anything for manual installation. It is just a [Gradle plugin](https://github.com/sherter/google-java-format-gradle-plugin). In `NullAway` it is in root [build.gradle](https://github.com/uber/NullAway/blob/master/build.gradle#L18). 

`Spotless` looks much more customizable though and is a superset of the Gradle plugin.

hzsweers:
yeah I think I'd rather spotless. This project doesn't use google's style guide anyway :). It's similar but with some changes (I don't like how GJF handles lambdas and always multilines field annotations)

VisheshVadhera:
@hzsweers @bangarharshit Can I open a PR for this one?

hzsweers:
I'll handle this one, I've bee prototyping it in another repo to iron out issues. Thanks though!

VisheshVadhera:
Sounds good!

VisheshVadhera:
@hzsweers Any other way by which I can contribute to this project?

hzsweers:
You could look at #181 if you want, just drop a comment first asking if anyone else is already working on it, and if no one responds by tomorrow, I'd say go for it :)

PR #310, opened 2018-12-12
Add overrideScopes flag to lint check

### LINKED ISSUE #298: Make types with scope additive by default, add flag to make it exclusive
I suspect most people just want to add custom scopes

- [ ] Lint
- [x] Error Prone

shaishavgandhi05:
Adds an `overrideScopes` flag that would exclude the default scopes in the lint check. 

Resolves #298

### DIFF
--- a/static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
+++ b/static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
@@ -109,7 +109,7 @@ class AutoDisposeDetector: Detector(), SourceCodeScanner {
 
   override fun beforeCheckRootProject(context: Context) {
     var overrideScopes = false
-    val scopes = HashSet<String>()
+    val scopes = mutableSetOf<String>()
 
     // Add the custom scopes defined in configuration.
     val props = Properties()
--- a/static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
+++ b/static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
@@ -48,6 +48,7 @@ import java.util.EnumSet
 
 internal const val CUSTOM_SCOPE_KEY = "autodispose.typesWithScope"
 internal const val LENIENT = "autodispose.lenient"
+internal const val OVERRIDE_SCOPES = "autodispose.overrideScopes"
 
 /**
  * Detector which checks if your stream subscriptions are handled by AutoDispose.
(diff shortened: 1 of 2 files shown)


static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -107,8 +108,8 @@ class AutoDisposeDetector: Detector(), SourceCodeScanner {
   private var lenient: Boolean = false
 
   override fun beforeCheckRootProject(context: Context) {
-    // Add the default Android scopes.
-    val scopes = HashSet(DEFAULT_SCOPES)
+    var overrideScopes = false
+    val scopes = HashSet<String>()
```
hzsweers: nit: can we make this a more idiomatic `mutableSetOf()` here while we're at it?

static-analysis/autodispose-lint/src/test/kotlin/com/uber/autodispose/lint/AutoDisposeDetectorTest.kt
```diff
@@ -530,6 +530,57 @@ class AutoDisposeDetectorTest {
         .expectClean()
   }
 
+  @Test fun overrideCustomScopeWithoutAutoDispose() {
+    val properties = projectProperties()
+    properties.property(CUSTOM_SCOPE_KEY, "com.uber.autodispose.sample.ClassWithCustomScope")
+    properties.property(OVERRIDE_SCOPES, "true")
+    properties.to(AutoDisposeDetector.PROPERTY_FILE)
+
+    lint().files(rxJava2(), LIFECYCLE_OWNER, ACTIVITY, properties, kotlin("""
```
hzsweers: nit - for readability, please put each of these args starting on separate lines

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
shaishavgandhi05: d476f27e99b04f6d1fe8b4c782914292811689a6

static-analysis/autodispose-lint/src/test/kotlin/com/uber/autodispose/lint/AutoDisposeDetectorTest.kt
shaishavgandhi05: 3c42327dcf482d18b76ed02b31e491fcf282fa52

Have changed it for the two methods I added but the remaining ones have that. Can do another PR for those to clean up the formatting there. On a side note, should we think about a checkstyle to enforce conventions?

hzsweers: cleaning up the rest separately is fine. There's an open issue for code style #154, but in general try to match what the rest of the project does too. My goal would be to use something like spotless

PR #316, opened 2018-12-13
Rename static analysis check name

shaishavgandhi05:


shaishavgandhi05:
Have already updated the wiki

PR #318, opened 2018-12-17
Update RxLifecycle to 3.0.0

shaishavgandhi05:
No breaking changes except the AndroidX change. [Changelog here](https://github.com/trello/RxLifecycle/blob/master/CHANGELOG.md#300-2018-10-25)


hzsweers:
These are all breaking changes with the package name change. I think we need to have an rxlifecycle3-interop artifact instead

shaishavgandhi05:
Hmmm yeah that's true. Good point.

Issue #302, opened 2018-12-03
Suggested fixes in static analysis


ZacSweers:
Currently, we don't offer a suggested fix in the checkers. We _could_ potentially offer one though, as we do know the type that's requesting the scope, and could just always suggest something like

```java
Observable.just(1)
    .as(autoDisposable(DetectedScopedClass.this)) // Suggest this
    .subscribe()
```

Potential issues:
* `DetectedScopedClass.this` might be overly pedantic when it could just be `this` in most cases
* In custom scoped classes, the mechanics may be different
  * A prime example of this is RIB Workers: https://github.com/uber/RIBs/blob/master/android/libraries/rib-base/src/main/java/com/uber/rib/core/Worker.java
* We want humans to review the fix, and at least error prone's is very easy to just sort of yolo-accept the suggestion without thinking about it

On the face of it, I think these might do more harm than good. If it was more like a templated approach like structural replace, that could be different (but likely only lint-supported in the IDE)

shaishavgandhi05:
Nice! Some additional thoughts:

* `DetectedScopedClass.this` might also be risky for Kotlin classes. We would have to do the Kotlin-esque `this@DetectedScopedClass`. In Lint, I believe we can find out if the class is Kotlin or Java so we can do this.
* For the Android scopes, we'll have to give suggestions like `AndroidLifecycleScopeProvider.from(this)`

hzsweers:
Let's punt on this for now. Not being able to discern anything but trivial cases gives me pause on trying to do something like this

ZacSweers:
Going to close this out since there's not a good path forward on this, happy to revisit if more proposals come up

ShaishavGandhi:
I wanted to reopen this since I think we're in a better position to execute on this now. Since we closed on the issue, we've merged the ktx artifacts into the main artifacts. This lets us confidently suggest quick fixes for Java and Kotlin files. 

@ZacSweers What do you think?

ZacSweers:
The original issue wasn't artifacts being split, rather this bit

> Let's punt on this for now. Not being able to discern anything but trivial cases gives me pause on trying to do something like this

Which I don't think has changed now

ShaishavGandhi:
I think we can cover the following cases:
* When a component that contains the subscription implements ScopeProvider, we can suggest auto-fixes for both Java and Kotlin file with `as(autoDisposable(this)` or `autoDisposable(this)` 
[...]

PR #321, opened 2018-12-19
PoC - Lint Quick Fix

shaishavgandhi05:
PoC for the lint part in #302. 

This demonstrates how the lint fix would work for Java files. Right now it only works on subscriptions within AutoDispose scope. The reason for only supporting Java files is uncertainty about whether for Kotlin files, ktx artifacts will be included. However, it is certainly possible to go ballistic and check if ktx artifacts are present (lint allows you to check dependencies, though I don't know whether those are deep dependencies). 

autodispose_replace

Can add more tests with other scenarios like subscribeWith etc. to demonstrate how the quick fix looks there.


hzsweers:
I've been thinking a lot about this, and I think that we shouldn't do this for now. It's one thing to look at something and see if it's obviously wrong, but I think it's much harder to deduce a fix that's always right.

shaishavgandhi05:
That's fair. Seeing as to how limited the functionality is turning out to be, combined with so many possibilities (Kotlin vs Java, ktx vs core artifacts, AutoDispose scopes vs Android scopes), it's best to revisit at another time.

Issue #296, opened 2018-12-03
Update to Gradle 5.0


ZacSweers:
Requires a newer version of AGP first

Working branch: https://github.com/uber/AutoDispose/tree/z/gradle5

PR #326, opened 2019-01-15
Upgrade to Gradle Plugin 3.3.0

shaishavgandhi05:


sample/src/main/res/layout/activity_disposing_viewmodel.xml
```diff
@@ -27,6 +27,7 @@
       android:layout_width="match_parent"
       android:layout_height="wrap_content"
       android:text="@string/disposing_viewmodel_activity_message"
+      app:layout_constraintTop_toTopOf="parent"
```
shaishavgandhi05: Fixed a lint error that was caught after moving to 3.3.0

shaishavgandhi05:
This also unblocks #296

PR #329, opened 2019-01-23
Upgrade to Kotlin 1.3.20!

shaishavgandhi05:
Faster builds with this one!

~`./gradlew clean check` previously took 1m 32s. Down to 52-53s.~
Had to disable the parallel build. It caused connectedTests to fail.


ZacSweers:
Interesting. Any chance you could take a before/after build scan?

shaishavgandhi05:
Yeah it's mainly because of parallelization of Gradle.

PR #335, opened 2019-03-18
Refer to main gradle.properties for lint configuration

### LINKED ISSUE #333: Lint error with lenient mode activated
Hi,
I wanted to handle myself the lifecycle of this request, so I removed my Autodispose call but the lenient error appeared.
I do have activated lenient mode in my gradle.properties, cleaned and rebuild, still the error remains.

**Library version**: `1.1.0`

**Repro steps or stacktrace**:

In this case, disposables is a CompositeDisposable which I'm disposing in the onDetach function

shaishavgandhi05:
**Description**: This changes the lint check to look for app level `gradle.properties` file instead of every library's. Relevant [lint thread](https://groups.google.com/forum/#!topic/lint-dev/fyY0Yz_n_Rk). This would require that you app level `build.gradle` has `checkDependencies=true` in it's `lintOptions`. Only then will lint be able to figure out that the library is part of the main project.

**Related issue(s)**: Resolves #333

### DIFF
--- a/sample/build.gradle
+++ b/sample/build.gradle
@@ -36,6 +36,10 @@ android {
   sourceSets {
     main.java.srcDirs += 'src/main/kotlin'
   }
+
+  lintOptions {
+    checkDependencies true
+  }
 }
 
 dependencies {
--- a/sample/gradle.properties
+++ b/sample/gradle.properties
@@ -1,1 +1,2 @@
 autodispose.typesWithScope=com.uber.autodispose.sample.CustomScope
+autodispose.lenient=false
--- a/static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
+++ b/static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
@@ -114,7 +114,7 @@ class AutoDisposeDetector: Detector(), SourceCodeScanner {
 
     // Add the custom scopes defined in configuration.
     val props = Properties()
-    context.project.propertyFiles.find { it.name == PROPERTY_FILE }?.apply {
+    context.mainProject.propertyFiles.find { it.name == PROPERTY_FILE }?.apply {
       val content = StringReader(context.client.readFile(this).toString())
       props.load(content)
       props.getProperty(CUSTOM_SCOPE_KEY)?.let { scopeProperty ->


sample/build.gradle
```diff
@@ -36,6 +36,10 @@ android {
   sourceSets {
     main.java.srcDirs += 'src/main/kotlin'
   }
+
+  lintOptions {
+    checkDependencies true
```
ZacSweers: What does this do exactly?

shaishavgandhi05: For Lint to detect the `mainProject`, you'd need your `:app` module to have `checkDependencies = true`. This basically means that when you do `./gradlew :app:lint`, it goes down and runs lint on its dependencies also. 

From Tor's [message](https://groups.google.com/forum/#!searchin/lint-dev/checkDependencies%7Csort:date/lint-dev/RGTvK_uHQGQ/5ie9w6nTBAAJ)

> ---------------------------------------
> Only analyze app/leaf modules
> ---------------------------------------
> 
> If you have divided your project into many smaller modules - a number of libraries and just a couple of app modules, it's much better to (a) turn on checkDependencies mode and (b) only run lint on the app modules, instead of recursively running lint on each module.
> 
> To do this, first add this to your app module's build.gradle file:
> ```
> android {
>     ...
>     lintOptions {
>         ...
>         checkDependencies true
>         ...
>     }
> }         
> ```
> 
> Then instead of running ./gradlew lintDebug, run ./gradlew :app:lintDebug.
> 
> Since you've turned on check-dependencies-mode, running lint on the app module will also run it on all the dependent modules the app depends on -- e.g. all the libraries. But this should also make things run a lot faster, since it's a single lint invocation, where lint shares a lot more computation (such as symbol resolution in the SDK and various shared libraries, etc).  
> 
> This isn't just a good idea from a performance perspective; you'll also get more accurate results. For example, the unused resource analysis will now correctly know whether a resource defined in a library is consumed by the app module; that doesn't happen when you run lint first on the library, and later on just the app, which is what happens when you run ./gradlew lintDebug. As a bonus you'll also get a single HTML report containing all the results from your codebase, instead of having to hunt around the tree for all the various reports and look at each one separately.


For AutoDispose, this is required since `mainProject` will only refer to the parent project when you have `checkDependencies=true`. If you run lint individually on a library module, `mainProject` = `project` and there's no way to get access to parent project.

ZacSweers: Interesting. I always thought this would be slower for some reason

PR #336, opened 2019-03-18
Add animal-sniffer to guard against Java 8 only APIs

shaishavgandhi05:
I was looking at integrating Animal Sniffer elsewhere and I found that the integration in AutoDispose was non-existent. In #264, I added a Java 8 API to show that Animal Sniffer works. Then when I was supposed to revert the Java 8 API commit, I reverted actual animal sniffer integration. 

source

So now, this build will fail because the Java 8 API is still existent (it didn't harm anyone since it was only a test that wasn't used by anyone). I'll remove it once this build fails on CI.


shaishavgandhi05:
[This build](https://travis-ci.org/uber/AutoDispose/builds/508011608?utm_source=github_status&utm_medium=notification) fails coz of Animal Sniffer. Reverting the offending API now.

Issue #340, opened 2019-04-05
Move remaining -ktx artifacts into mainline repo


ZacSweers:
Following the same pattern as https://github.com/uber/AutoDispose/pull/339

- [ ] `autodispose-android-archcomponents-ktx`
- [ ] `autodispose-android-archcomponents-test-ktx`
- [x] `autodispose-android-ktx`
- [ ] `autodispose-lifecycle-ktx` (this is deprecated already but let's delete the artifact)

PR #341, opened 2019-04-06
Fold autodispose-android-ktx into autodispose-android

shaishavgandhi05:
Ref #340


README.md
```diff
@@ -186,9 +186,6 @@ There are three artifacts with extra support for Android:
 * `autodispose-android-archcomponents` has a `AndroidLifecycleScopeProvider` for use with `LifecycleOwner` and `Lifecycle` implementations.
 * `autodispose-android-archcomponents-test` has a `TestLifecycleOwner` for use in testing.
 
-For each artifact, there is a corresponding kotlin extensions artifact with it. Example:
```
ZacSweers: Let's tweak this to just say it's WIP for next release and link the ticket

android/autodispose-android/src/main/java/com/uber/autodispose/android/AutoDisposeAndroid.kt
```diff
@@ -15,6 +15,7 @@
  */
 
 @file:Suppress("NOTHING_TO_INLINE")
+@file:JvmName("AutoDisposeAndroid")
```
ZacSweers: let's nix this and call it KotlinExtensions like the main artifact

shaishavgandhi05: da380b7f7b50d2c390ab2beab8551683f79c8d28

shaishavgandhi05:
Still trying to figure out why the `androidJavadoc` is being run instead of dokka 😪

README.md
shaishavgandhi05: 6e96d95

shaishavgandhi05:
@ZacSweers Can you take a look again? It's been a while so might be good to get a fresh eye again.

build.gradle
```diff
@@ -95,7 +95,7 @@ subprojects {
   boolean isLint = project.path.endsWith("-lint")
   boolean isKotlin = project.path.endsWith("-ktx") || isLint || isMixedSourceSet
   boolean isSample = project.name == "sample"
-  boolean isJavaLibrary = !isAndroidLibrary && !isKotlin && !isSample || isMixedSourceSet
+  boolean isJavaLibrary = !isAndroidLibrary && !isKotlin && !isSample || (isMixedSourceSet && !isAndroidLibrary)
```
ZacSweers: > (isMixedSourceSet && !isAndroidLibrary)

Isn't `!isAndroidLibrary` redundant?

README.md
```diff
@@ -186,6 +186,8 @@ There are three artifacts with extra support for Android:
 * `autodispose-android-archcomponents` has a `AndroidLifecycleScopeProvider` for use with `LifecycleOwner` and `Lifecycle` implementations.
 * `autodispose-android-archcomponents-test` has a `TestLifecycleOwner` for use in testing.
 
+Kotlin extension artifacts are a [WIP for the next release](https://github.com/uber/AutoDispose/issues/340).
```
ZacSweers: Generally shouldn't mark WIP stuff in the README, let's remove this for now. README should just reflect current master

build.gradle
shaishavgandhi05: The OR branch is always evaluated and if it's an Android library, we wouldn't want to apply the 'java-library` plugin (it gives an error).

README.md
shaishavgandhi05: Makes sense d12ba8c3c8f7a1e622cbe58fb887827a8134b0af

ZacSweers:
@shaishavgandhi05 do you know what `continuous-integration/travis-ci` is? It should only be running the PR build. I can't see any information about it either, and it appears to be hung up. We'll have to force land these as admin due to it never reporting back.

@briankhsieh maybe is this something new from the org side relating to the travis-ci.com migration?

PR #343, opened 2019-04-19
Update to AGP 3.4.0

### LINKED ISSUE #325: New lintPublish configuration
The names in AGP 3.3 have changed a bit: https://developer.android.com/studio/releases/gradle-plugin

New lint check configuration:

> **lintChecks**: This is an existing configuration that you should use only for lint checks you want to run when building your project locally.
> **lintPublish**: Use this new configuration for library projects in which you want to include your lint checks in the published AAR. This means that projects that consume your library also apply those lint checks.
> Note: Due to this behavior change, existing library modules that upgrade to Android Gradle plugin 3.3.0 and higher need to use the new lintPublish configuration to include custom lint checks in the published AAR.

shaishavgandhi05:
Resolves #325 

One caveat: The `autodispose` module (since it's a Java module) doesn't have `lintPublish` configuration. I've filed a bug in the [Issue Tracker](https://issuetracker.google.com/issues/130837765).

### DIFF
--- a/android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/LifecycleEventsObservable.java
+++ b/android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/LifecycleEventsObservable.java
@@ -15,7 +15,7 @@
  */
 package com.uber.autodispose.android.lifecycle;
 
-import static androidx.annotation.RestrictTo.Scope.LIBRARY;
+import static androidx.annotation.RestrictTo.Scope.LIBRARY_GROUP;
 import static androidx.lifecycle.Lifecycle.Event.ON_CREATE;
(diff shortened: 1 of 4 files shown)


android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/LifecycleEventsObservable.java
```diff
@@ -34,7 +34,7 @@
 import io.reactivex.Observer;
 import io.reactivex.subjects.BehaviorSubject;
 
-@RestrictTo(LIBRARY)
+@RestrictTo(LIBRARY_GROUP)
```
shaishavgandhi05: This throws a lint error now. `LIBRARY_GROUP` seems correct since we're still restricting it to `com.uber.autodispose` gradle id.

Issue #262, opened 2018-10-10
Fix publishDocs task


ZacSweers:
Two parts

- Need a new release of osstrich for new mavencentral UI changes
- Something changed in gradle and we have to specify repos now

```
* What went wrong:
Execution failed for task ':publishDocs'.
> Could not resolve all files for configuration ':osstrich'.
   > Cannot resolve external dependency com.squareup.osstrich:osstrich:1.3.0 because no repositories are defined.
     Required by:
         project :
```

shaishavgandhi05:
Fixed in #345

PR #345, opened 2019-05-08
Update gradle and osstrich

ZacSweers:
Osstrich now works again!


shaishavgandhi05:
I'm guessing this resolves #262 ?

ZacSweers:
Yep!

Issue #347, opened 2019-05-12
Add more top-level extensions for scopes


ZacSweers:
Right now we add scope extensions for scopable types, but let's go the whole 9 yards and add overloads for the common cases to save folks time

so instead of just offering

```kotlin
observable.autoDisposable(AndroidLifecycleScopeProvider.from(this)).subscribe()
```

Let's do this too

```kotlin
observable.autoDisposable(this).subscribe()
```

And do the same for other extension points. Only concern I'd have is possibly conflicting imports if the extensions are in different packages (but easy to test that).

shaishavgandhi05:
Can also add an overload for the `untilEvent`:
```kotlin
observable.autoDisposable(this, ON_DESTROY).subscribe()
```

ZacSweers:
Yep! And can make it have a default value too

shaishavgandhi05:
Resolved in #348

PR #348, opened 2019-05-13
Add additional extension methods for android-archcomponents

shaishavgandhi05:
Part of #347 

I've added an additional `TestKotlinActivity` in the `sample` just to test out the compilation of extension methods.


android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/KotlinExtensions.kt
```diff
@@ -77,3 +90,93 @@ inline fun Lifecycle.scope(
...
+@CheckReturnValue
+inline fun <T> Flowable<T>.autoDisposable(lifecycleOwner: LifecycleOwner): FlowableSubscribeProxy<T> =
+    this.`as`(AutoDispose.autoDisposable(AndroidLifecycleScopeProvider.from(lifecycleOwner)))
+
+/**
+ * Extension that proxies to [Flowable.as] + [AutoDispose.autoDisposable] and takes an [untilEvent] when
+ * subscription will be disposed.
+ */
+@CheckReturnValue
+inline fun <T> Flowable<T>.autoDisposable(lifecycleOwner: LifecycleOwner, untilEvent: Event): FlowableSubscribeProxy<T> =
```
ZacSweers: Let's default the untilEvent to something sensible and then just have one overload?

shaishavgandhi05: Can we really default an `untilEvent` reliably for everyone? Usually the `untilEvent` is inferred from when in the lifecycle you subscribe. However that inference logic is private in `AndroidLifecycleScopeProvider`.

ZacSweers: I'd be fine with it package private

shaishavgandhi05:
> Why not put this in the sample?
Thought about it but I meant to keep that file as a dumping ground for testing Kotlin compatibility. When I add the `View` extensions, that can also be tested in the same file. Something on the lines on [KotlinSourceCompatibilityTest](https://github.com/square/okhttp/blob/6c13e6bd5df94ba95e008d477fe17c7a6b13a8fa/okhttp/src/test/java/okhttp3/KotlinSourceCompatibilityTest.kt) from OkHttp.

ZacSweers:
Works for me!

android/autodispose-android-archcomponents/src/androidTest/java/com/uber/autodispose/android/lifecycle/KotlinExtensionsTest.kt
```diff
@@ -0,0 +1,323 @@
...
+import io.reactivex.subjects.PublishSubject
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class KotlinExtensionsTest {
+
+  companion object {
+    private val LOGGER = RecordingObserver.Logger {
+      message -> Log.d(KotlinExtensionsTest::class.java.simpleName, message) }
```
ZacSweers: nit: closing bracket on next line

android/autodispose-android-archcomponents/src/androidTest/java/com/uber/autodispose/android/lifecycle/KotlinExtensionsTest.kt
```diff
@@ -0,0 +1,323 @@
...
+import com.google.common.truth.Truth.assertThat
+import com.uber.autodispose.OutsideScopeException
+import com.uber.autodispose.android.lifecycle.test.TestLifecycleOwner
+import com.uber.autodispose.test.RecordingObserver
+import io.reactivex.subjects.PublishSubject
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class KotlinExtensionsTest {
```
shaishavgandhi05: These tests are basically copied over from `AndroidLifecycleScopeProviderTest`. I wanted to verify behavior and make sure the extensions don't behave differently. I can consolidate the two if the refactor looks ok

android/autodispose-android-archcomponents/src/androidTest/java/com/uber/autodispose/android/lifecycle/KotlinExtensionsTest.kt
```diff
@@ -0,0 +1,323 @@
...
+    val t = o.takeError()
+    assertThat(t).isInstanceOf(IllegalStateException::class.java)
+    assertThat(t.message).contains("main thread")
+    o.assertNoMoreEvents()
+    assertThat(d.isDisposed).isTrue()
+  }
+
+  @Test(expected = OutsideScopeException::class)
+  @UiThreadTest
+  fun observable_offAfterDestroy_shouldFail() {
```
shaishavgandhi05: This test is the only one where the behavior is different. 
Since we're trying to get the corresponding `Lifecycle.Event` to the `Lifecycle.State` super early (while trying to resolve the default `untilEvent`), during `ON_DESTROY` the `OutsideScopeException` will be called immediately.

android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/KotlinExtensions.kt
shaishavgandhi05: ec8cd98

android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/internal/CorrespondingEventsUtil.java
```diff
@@ -0,0 +1,71 @@
...
+          case ON_PAUSE:
+            return Lifecycle.Event.ON_STOP;
+          case ON_STOP:
+          case ON_DESTROY:
+          default:
+            throw new LifecycleEndedException("Lifecycle has ended! Last event was " + lastEvent);
+        }
+      };
+
+  public static Lifecycle.Event getCorrespondingEvent(Lifecycle lifecycle) {
```
ZacSweers: nit: this name is probably more accurate as something like `getNearestPreviousEvent`

android/autodispose-android-archcomponents/src/androidTest/java/com/uber/autodispose/android/lifecycle/KotlinExtensionsTest.kt
ZacSweers: hmm, can we defer the resolution of `untilEvent` through something like `UntilEventFunction` and `CorrespondingEventsFunction` in the existing java code? I'd be fine making that 

Hadn't seen this before approving, let's try to keep the behavior the same before merging

shaishavgandhi05: That's the part that is tricky (and probably not possible). `untilEvent` has to be resolved immediately since it has a default value. In our case it becomes a conundrum because we need it to be deferred.

sample/src/androidTest/java/com/uber/autodispose/TestKotlinActivity.kt
```diff
@@ -31,7 +33,7 @@ import java.util.concurrent.TimeUnit
  * Test Activity class to verify compilation of various extension functions.
  */
 @Ignore("Since it's only used to verify compilation of the extension functions")
-class TestKotlinActivity : AppCompatActivity() {
+class TestKotlinActivity : AppCompatActivity(), ScopeProvider {
```
ZacSweers: Just to doubly confirm - it doesn't get confused about ambiguous types here if targeting this now?

android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/KotlinExtensions.kt
```diff
@@ -91,92 +91,74 @@ inline fun Lifecycle.scope(
...
-/**
- * Extension that proxies to [Flowable.as] + [AutoDispose.autoDisposable]
- */
-@CheckReturnValue
-inline fun <T> Flowable<T>.autoDisposable(lifecycleOwner: LifecycleOwner): FlowableSubscribeProxy<T> =
-    this.`as`(AutoDispose.autoDisposable(AndroidLifecycleScopeProvider.from(lifecycleOwner)))
-
 /**
  * Extension that proxies to [Flowable.as] + [AutoDispose.autoDisposable] and takes an [untilEvent] when
  * subscription will be disposed.
```
ZacSweers: nit: doc the parameters

android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/KotlinExtensions.kt
```diff
@@ -91,92 +91,74 @@ inline fun Lifecycle.scope(
...
-    this.`as`(AutoDispose.autoDisposable(AndroidLifecycleScopeProvider.from(lifecycleOwner, untilEvent)))
-
-/**
- * Extension that proxies to [Observable.as] + [AutoDispose.autoDisposable]
- */
-@CheckReturnValue
-inline fun <T> Observable<T>.autoDisposable(lifecycleOwner: LifecycleOwner): ObservableSubscribeProxy<T> =
-    this.`as`(AutoDispose.autoDisposable(AndroidLifecycleScopeProvider.from(lifecycleOwner)))
+inline fun <T> Flowable<T>.autoDisposable(lifecycleOwner: LifecycleOwner, untilEvent: Event? = null): FlowableSubscribeProxy<T> {
+  untilEvent?.let {
```
ZacSweers: nit: let's just keep this a simple `if (untilEvent == null)` and extract the `return` to the top with the if statement as the expression

ZacSweers:
Couple nits but LGTM!

sample/src/androidTest/java/com/uber/autodispose/TestKotlinActivity.kt
shaishavgandhi05: No, it doesn't. Basically, it will always just resolve to one of those extension functions. It will ask you for one of them. If you're overloading with the `untilEvent`, it will always resolve to the one with `LifecycleOwner`.

android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/KotlinExtensions.kt
shaishavgandhi05: That's what I had before but then I was like maybe I should be more Kotlin-y 😄

shaishavgandhi05: 38de073

android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/KotlinExtensions.kt
shaishavgandhi05: 38de073

PR #362, opened 2019-07-11
Finalize website assets

ShaishavGandhi:
Added the new fonts locally and updated the `RELEASING.md`.


ShaishavGandhi:
> Is it not possible to link from a remote point for fonts?

Had an email chain internally and couldn't get a cdn link. The recommendation was to include this font in the repo itself. 

> Also can you do a separate write-up (blog or otherwise) for actually updating and maintaining these docs beyond just publishing steps?

Yup, can do that in a follow up.

Issue #220, opened 2018-07-04
Investigate alternative CI solutions


ZacSweers:
Travis takes so long for android tests, would like to explore this more.

shaishavgandhi05:
I use CircleCI in most of my projects and I find it great. It's really fast and configuration is simpler. 

One caveat is that CircleCI 2.0 doesn't really support [testing on emulator](https://support.circleci.com/hc/en-us/articles/360000028928-Testing-with-Android-emulator-on-CircleCI-2-0). They seem to be recommending using Firebase Test Lab.

hzsweers:
interesting... the lack of emulator support is frustrating :|, as that's the main thing that makes travis slow for this project.

Have any example configs you could point me to? Never used it personally, and our only other project using it (okbuck) doesn't have an android configuration for it afaik

shaishavgandhi05:
Here's a [configuration](https://github.com/shaishavgandhi05/navigator/blob/master/.circleci/config.yml) from one of my projects. It's pretty much what they recommend in their [guide](https://circleci.com/docs/2.0/language-android/#sample-configuration).

hzsweers:
Thanks!

PR #365, opened 2019-07-27
Move to CircleCI

ShaishavGandhi:
Still WIP
Ref #220


ZacSweers:
Does Uber have an FTL account that we can use for instrumentation tests?

ShaishavGandhi:
I saw some StackOverflow where they were able to run espresso tests on emulator without FTL. Seeing if that works out otherwise might have to go the FTL path.

ShaishavGandhi:
Have been able to get the emulator tests running! 
Going to take a look at uploading snapshots and then we can phase out Travis.

ShaishavGandhi:
I think I'm going to continue this on my fork and come back when I have something that works.

PR #372, opened 2019-09-07
Rename extension functions

### LINKED ISSUE #355: API proposal: improve kotlin extensions names
Right now the Kotlin extensions all are like this

```kotlin
fooObservable
    .autoDisposable(...)
    .subscribe()
```

I've been wondering about this for awhile from a readability standpoint. In Java it made more sense because it was with `as()`

```java
myObservable
    .as(autoDisposable(...))
    .subscribe()
```

But with kotlin, maybe we can make it flow better with a different name

```kotlin
fooObservable
    .disposeWith(...)
    .subscribe()
```

Other options - `disposeOn()`, `autoDisposeWith()`, `autoDisposeOn()`, `autoDispose()`, ???

We can leave the old APIs for binary compatibility and deprecate them with a simple replace and error severity for easy migration.

Would require updating the lint as well.

ShaishavGandhi:
Resolves #355 

I feel like we can keep the severity to be WARN for 1.4.0? That way the upgrade to 1.4.0 can be smooth while 2.0.0 can be where we break.

### DIFF
--- a/android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/KotlinExtensions.kt
+++ b/android/autodispose-android-archcomponents/src/main/java/com/uber/autodispose/android/lifecycle/KotlinExtensions.kt
@@ -99,7 +99,28 @@ inline fun Lifecycle.scope(
  * @param untilEvent Optional lifecycle event when subscription will be disposed.
  */
 @CheckReturnValue
+@Deprecated(
+    replaceWith = ReplaceWith("autoDispose(lifecycleOwner, untilEvent)"),
+    message = "Renamed to `autoDispose`",
+    level = DeprecationLevel.ERROR
+)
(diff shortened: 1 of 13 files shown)


ZacSweers:
The upgrade is the same regardless with the ReplaceWith, let's just go with hidden. Or does replacewith not work with hidden?

ShaishavGandhi:
Yeah, ReplaceWith doesn't work with HIDDEN. HIDDEN will give an error but not suggest the fix.

Making it ERROR will move consumers over immediately but then will feel like a breaking change. I think WARNING will work well for now, give users time to upgrade and with 2.0.0, they'll know the extensions won't exist anymore. 

What do you think?

ZacSweers:
I disagree. ReplaceWith exists to mitigate issues exactly like this with minimal friction for users, and this isn't a breaking change so we don't need to wait for 2.0 :)

PR #375, opened 2019-09-14
Add AutoDisposeContext + withScope functions

ZacSweers:
This allows for some more idiomatic kotlin usage like so:

```kotlin
withScope(sourceScope) {
  someObservable.autoDispose().subscribe()
}
```

### DIFF
--- a/autodispose/src/main/java/com/uber/autodispose/KotlinExtensions.kt
+++ b/autodispose/src/main/java/com/uber/autodispose/KotlinExtensions.kt
@@ -255,12 +255,12 @@ inline fun <T> ParallelFlowable<T>.autoDisposable(provider: ScopeProvider): Para
 inline fun <T> ParallelFlowable<T>.autoDispose(provider: ScopeProvider): ParallelFlowableSubscribeProxy<T> =
     this.`as`(AutoDispose.autoDisposable(provider))
 
-/** Executes a [body] with under an [AutoDisposeContext] backed by the given [scope]. */
+/** Executes a [body] with an [AutoDisposeContext] backed by the given [scope]. */
 fun withScope(scope: ScopeProvider, body: AutoDisposeContext.() -> Unit) {
   withScope(completableOf(scope), body)
 }
 
-/** Executes a [body] with under an [AutoDisposeContext] backed by the given [completableScope]. */
+/** Executes a [body] with an [AutoDisposeContext] backed by the given [completableScope]. */
 fun withScope(completableScope: Completable, body: AutoDisposeContext.() -> Unit) {
   val context = RealAutoDisposeContext(completableScope)
   context.body()
--- a/autodispose/src/main/java/com/uber/autodispose/KotlinExtensions.kt
+++ b/autodispose/src/main/java/com/uber/autodispose/KotlinExtensions.kt
@@ -18,6 +18,7 @@
 
 package com.uber.autodispose
 
+import com.uber.autodispose.Scopes.completableOf
 import io.reactivex.Completable
 import io.reactivex.Flowable
 import io.reactivex.Maybe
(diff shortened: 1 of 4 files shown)


autodispose/src/main/java/com/uber/autodispose/KotlinExtensions.kt
```diff
@@ -253,3 +254,48 @@ inline fun <T> ParallelFlowable<T>.autoDisposable(provider: ScopeProvider): Para
 @CheckReturnValue
 inline fun <T> ParallelFlowable<T>.autoDispose(provider: ScopeProvider): ParallelFlowableSubscribeProxy<T> =
     this.`as`(AutoDispose.autoDisposable(provider))
+
+/** Executes a [body] with under an [AutoDisposeContext] backed by the given [scope]. */
```
ShaishavGandhi: I guess you want `with` or `under`? i.e 
`Executes a [body] under an ....`
or
`Executes a [body] with an...`

autodispose/src/main/java/com/uber/autodispose/KotlinExtensions.kt
```diff
@@ -253,3 +254,48 @@ inline fun <T> ParallelFlowable<T>.autoDisposable(provider: ScopeProvider): Para
...
+}
+
+/** Executes a [body] with under an [AutoDisposeContext] backed by the given [completableScope]. */
+fun withScope(completableScope: Completable, body: AutoDisposeContext.() -> Unit) {
+  val context = RealAutoDisposeContext(completableScope)
+  context.body()
+}
+
+/**
+ * A context intended for use as `AutoDisposeCoroutineScope.() -> Unit` function body parameters
```
ShaishavGandhi: Given that this is in the `autodispose` module and that you can use it with a normal `ScopeProvider` as well, we can probably skip the reference to coroutines.

ShaishavGandhi:
How about adding convenience functions like:
```kotlin
fun withScope(coroutineScope: CoroutineScope, body: AutoDisposeContext.() -> Unit) {
  withScope(coroutineScope.toScopeProvider(), body)
}

```

autodispose/src/main/java/com/uber/autodispose/KotlinExtensions.kt
ZacSweers: That's a stale doc from the original PR, will fix

ZacSweers:
> How about adding convenience functions

I don't like it in this case tbh. I think convenience functions make sense for discoverability and shorthands for common cases, but that case isn't either of those.

autodispose/src/main/java/com/uber/autodispose/KotlinExtensions.kt
ZacSweers: AutoDisposeContext

autodispose/src/main/java/com/uber/autodispose/KotlinExtensions.kt
ZacSweers: 32e7e08

ShaishavGandhi:
Given that `withScope` is so similar to `withContext` from coroutines, it'd be perplexing to not overload `withScope` with a `CoroutineScope` since coroutines users will have an instinct of using the `with*` functions. Just my 2 cents.

ZacSweers:
withContext is not a scoping construct like this is though, whereas this is a scoping construct. I thought about calling this AutoDisposeScope at first, but context in this case in reference to the context that those extension functions are available. 

It's best not to think of this as having anything to do with coroutines or coroutine scope because it's unrelated functionality. The only thing in common really is that they use higher order functions with receivers :)

ZacSweers:
Also - I think a good API boundary here is supporting AutoDispose's core types, and anything else that can be interop'd back to them should just use that. Otherwise it's a slippery slop to supporting anything that Completable and ScopeProvider can interop out to: CoroutineScope, LifecycleProvider (2 and 3), etc. If someone wants to write their own extension function to get the API you're describing, great! But otherwise I feel pretty strongly about keeping the API surface area here down to just AutoDispose's own core components.

ShaishavGandhi:
> Also - I think a good API boundary here is supporting AutoDispose's core types, and anything else that can be interop'd back to them should just use that.

Yeah I strongly agree with that. Given the nature of composability of Rx, we'd end up supporting lots of overloads. 

The little inkling was mainly because of the similarity in names but it's only a minor inconvenience for people to just do `.asScopeProvider()`

PR #378, opened 2019-09-18
Support withScope() in AutoDisposeDetector

ZacSweers:
This adds support for the new `withScope()` API in `AutoDisposeDetector`. This works by detecting when that function is called, and then visits the contents of the lambda argument passed to it with the same machinery we used before. This refactors out that machinery a bit to allow defining a custom "isInScope" check, and in the case of `withScope()` it's just hardcoded to `true` when it checks if an enclosed subscribe() call is made in a scope

### DIFF
--- a/static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
+++ b/static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
@@ -26,7 +26,6 @@ import com.android.tools.lint.detector.api.JavaContext
 import com.android.tools.lint.detector.api.Scope
 import com.android.tools.lint.detector.api.Severity
 import com.android.tools.lint.detector.api.SourceCodeScanner
-import com.intellij.psi.PsiClass
 import com.intellij.psi.PsiMethod
 import com.intellij.psi.PsiType
 import com.intellij.psi.util.PsiUtil
@@ -87,6 +86,7 @@ class AutoDisposeDetector : Detector(), SourceCodeScanner {
     private const val SINGLE = "io.reactivex.Single"
     private const val MAYBE = "io.reactivex.Maybe"
     private const val COMPLETABLE = "io.reactivex.Completable"
+    private const val KOTLIN_EXTENSIONS = "com.uber.autodispose.KotlinExtensions"
 
     // The default scopes for Android.
     private val DEFAULT_SCOPES = listOf("androidx.lifecycle.LifecycleOwner",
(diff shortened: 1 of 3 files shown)


static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
```diff
@@ -143,19 +144,89 @@ class AutoDisposeDetector : Detector(), SourceCodeScanner {
...
-          evaluateMethodCall(node, it, context)
+        node.resolve()?.let { method ->
+          // Check if it's one of our withScope() higher order functions. If so, we handle that
+          // separately and visit the passed in lambda body and run the subscribe method call checks
+          // inside it with the "isInScope" check just hardcoded to true.
+          if (method.name == "withScope" &&
+              method.containingClass?.qualifiedName == KOTLIN_EXTENSIONS) {
+            val args = node.valueArguments
+            if (args.size == 2) {
+              val last = args[1]
```
ShaishavGandhi: Can also do `args.last()` i believe.

static-analysis/autodispose-lint/src/test/kotlin/com/uber/autodispose/lint/AutoDisposeDetectorTest.kt
```diff
@@ -1052,4 +1080,103 @@ class AutoDisposeDetectorTest {
...
+            kotlin("""
+          package foo
+          import io.reactivex.Completable
+          import io.reactivex.Observable
+          import com.uber.autodispose.withScope
+          class ExampleClass {
+            fun names() {
+              val observable = Observable.just(1)
+              withScope(Completable.complete()) {
+                observable.subscribe()
```
ShaishavGandhi: What happens if you have another line before this `subscribe()`? 

Let's say
```kotlin
val a = 1
observable.subscribe()
```

Just wondering if the `subscribe` has to be the only code in the LambdaExpression for lint to detect correctly.

static-analysis/autodispose-lint/src/main/kotlin/com/uber/autodispose/lint/AutoDisposeDetector.kt
ZacSweers: That does an extra size check, which we can skip in this case

static-analysis/autodispose-lint/src/test/kotlin/com/uber/autodispose/lint/AutoDisposeDetectorTest.kt
ZacSweers: Still works! The visitor will visit it like any other `UBlockExpression`. This is similar to if, say, lint runs on the contents of a `run { }` block or function body.

ShaishavGandhi: Cool, sounds good! Let's merge!

PR #381, opened 2019-09-18
Migrate to RxJava 3

### LINKED ISSUE #359: RxJava 3 Support
RxJava 3 is planned for later this year, but will use the same maven coordinates and package name. There is one significant change that affects AutoDispose: the `as()` operator has been folded back into `to()`, where the original `to()` overloads accepting `Function`s has been replaced with `as()`'s use of custom converter interfaces.

For Java consumers - this basically means no change is required from the AutoDispose side. Consumers would need to switch to `to()`, but all the static `autoDisposable()` APIs would work out of the box with that in RxJava 3. All we need to do is update docs for usage in both.

This does present an interesting problem for the kotlin extensions though, which all call through to `.as()` under the hood and would no longer function in RxJava 3.

My current thoughts right now are to try to avoid requiring a V2 for AutoDispose, and instead make RxJava 3 support somehow opt-in. An idea for this is to wrap support behind an abstraction in an rx3-only module compiled against rxjava 3. Then you can opt in the kotlin extensions to this via new AutoDispose plugin, and under the hood it would switch to use the rx3 support.

The rx3 support module would have something like this 
```kotlin
// Ugly name but we can't mark this as internal since it's another module, but maybe we can make it harder to discover
fun <T> Observable<T>.rx3toInteropForAutoDispose(converter: ObservableConverter<T, ObservableSubscribeProxy<T>>): ObservableSubscribeProxy<T> = to(converter)
```

Then in the main kotlin extensions we'd do something like this

```kotlin
fun <T> Observable<T>.autoDisposable(scope: ScopeProvider): ObservableSubscribeProxy<T> {
  val converter = AutoDispose.autoDisposable<T>(scope)
  return if (AutoDisposePlugins.useRx3()) {
    rx3toInteropForAutoDispose(converter)
  } else {
    `as`(converter)
  }
}
```

We'd need to add some sort of proguard warning suppression for consumers not using rx3 to safely ignore it missing from the classpath

ZacSweers:
Resolves #359


ShaishavGandhi:
We'll want a 2.x and 1.x branch right? So that we can keep supporting both RxJava2 and RxJava3 if there are any bug fixes.

ZacSweers:
Nope, I don't think that's necessary. We can backport bugfixes via branch
based on the last 1.x tag. RxJava did this because of a long tail of
concurrent maintenance and development, but we're not going to develop 1.x
any further.

ZacSweers:
Yep one step at a time

PR #386, opened 2019-09-24
Remove rxlifecycle interop

ZacSweers:
We can re-add if there's an rxlifecycle 4 with rxjava3 support


settings.gradle
```diff
@@ -62,8 +62,6 @@ if (System.getenv("ANDROID_HOME") != null) {
 include ':autodispose'
 include ':autodispose-coroutines-interop'
 include ':autodispose-lifecycle'
-//include ':autodispose-rxlifecycle'
```
ShaishavGandhi: Remove?

ZacSweers: They are removed locally... ? 🤔

ZacSweers: Yeah they are removed o_O

ShaishavGandhi: Nvm i misread 🤦‍♂

PR #388, opened 2019-09-24
Rename android arch lifecycle artifacts to androidx lifecycle naming

### LINKED ISSUE #379: 2.x: Rename android arch components artifacts
These can be renamed to just `autodispose-android-lifecycle` since they're really just androidx lifecycle extensions

ZacSweers:
Resolves #379 

Will merge manually to best-effort protect git history


android/autodispose-androidx-lifecycle-test/gradle.properties
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-POM_NAME=AutoDispose (Android Architecture Components Extensions)
-POM_ARTIFACT_ID=autodispose-android-archcomponents
+POM_NAME=AutoDispose (androidx-lifecycle test extensions)
```
ShaishavGandhi: My capitalizing OCD has trouble with `test extensions`

ZacSweers: Sorry, went for consistency here

ZacSweers: "Androidx-Lifecycle" would be more wrong

ShaishavGandhi: Yeah i know. Not suggesting we change, but just had to point out :D

PR #392, opened 2019-10-13
Make README just point to website

ZacSweers:
This allows us to version our docs and also dedupe maintenance of them


ShaishavGandhi:
Actually hold on. The current `deploy_website.sh` simply copies the README.md into the `docs` folder. If we're deleting the README, we should add a proper file in `docs` folder that will represent current state of AutoDispose.

ZacSweers:
Can we just make that the index.md?

ShaishavGandhi:
Yup that's what i meant. The script copies it as `index.md` in `docs` folder but doesn't check it in version control (it's added to gitignore). 

So we should copy it as `index.md` and then remove the gitignore and keep that as our versioned doc

ZacSweers:
Kk will update

PR #398, opened 2019-11-09
Test Patch

ShaishavGandhi:


ShaishavGandhi:
Yeah doesn't seem to trigger from my fork

ZacSweers:
Want to make this the PR that enables this on PRs? My main concern is accidentally doubling builds

ShaishavGandhi:
Yeah let me make a PR for that and see if it runs twice

ZacSweers:
I meant this one, but new one works too ha

PR #401, opened 2019-11-12
Remove executable bit from non-executable files

friederbluemle:
A bunch of non-executable files in the repo erroneously had the executable bit set. This fixes is by changing the file mode from `755` to `644`.


ZacSweers:
Well that's interesting! Shouldn't `gradlew.bat` keep its executable status?

friederbluemle:
> Well that's interesting! Shouldn't `gradlew.bat` keep its executable status?

Good question: No, `gradlew.bat` is used on Windows only, and Windows does not use POSIX file permissions. In any default setup (through Android Studio, or Gradle command line), _only_ `gradlew`, _not_ `gradlew.bat` has the executable bit set.

PR #402, opened 2019-11-12
Downgrade Gradle wrapper to 5.6.4

### LINKED ISSUE #400: Lower gradle version back to 5.x
AARs don't seem to be consumable when built with 6.x is: https://issuetracker.google.com/issues/144240061

friederbluemle:
Downgrades Gradle to the latest 5.x release.

Fixes #400

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.0-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists


ZacSweers:
Did this not make any changes to the gradle wrapper jar?

friederbluemle:
The update to `gradle-wrapper.properties` was done through the Gradle `wrapper` task (not by manually modifying the file). Since there was no change to the jar file I assume no changes.

For several version now, Gradle has fixed the ever changing jar file issue (likely caused by some file that had only a text timestamp updated). The contents of the jar file (just like the wrapper scripts) should only change very rarely.

ZacSweers:
I get stale state locally when I do this.

image


I'll take this over in a separate PR anyway as I want to add a validation layer from this https://gist.github.com/ZacSweers/91597b19744f5fe314313ae9b11d6879

ZacSweers:
Continued in #403

friederbluemle:
What do you mean by "stale state locally" - Which command did you run before `git status`?

ZacSweers:
The wrapper task _only_ updates the properties file. Subsequent Gradle runs
actually perform the bootstrapping, it’s weird

PR #407, opened 2019-11-21
Bring back Travis CI

ShaishavGandhi:
Hopefully temporarily while Uber ponders upgrading the paid plan. 

For now, can keep these two side-by-side


ShaishavGandhi:
Done!

PR #415, opened 2020-01-01
Update with proper emulator workaround

ShaishavGandhi:
Trying to see if this workaround for the emulator issue actually works.


ShaishavGandhi:
This is failing on current master as well so looks like it's because of RxAndroid's new snapshot.

PR #419, opened 2020-02-03
Android ViewModel ScopeProvider created

eugene-nefedov:
**Description**:
I found very convenient to use autoDispose() on a ViewModel without extending it from the AutoDisposeViewModel. 
So, I propose the next idea:
  1. Do not force the user to extend AutoDisposeViewModel
  2. Make it possible to work with direct ViewModel children

The implementation of idea is new ScopeProvider creation. Such a ScopeProvider binds to the ViewModel's lifecycle (in general meaning). The binding is done with the support of the Kotlin Coroutine Scope for ViewModel, which is provided as extension property ViewModel.viewModelScope in the androidx.lifecycle:lifecycle-viewmodel-ktx library.

If you find this idea is interesting too, feel free to tell me what is missing in my PR.

### DIFF
--- a/android/autodispose-androidx-lifecycle/build.gradle
+++ b/android/autodispose-androidx-lifecycle/build.gradle
@@ -24,6 +24,7 @@ dependencies {
 
   implementation project(':android:autodispose-android')
   implementation deps.rx.android
+  implementation deps.androidx.lifecycle.ktx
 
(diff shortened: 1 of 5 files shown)


ZacSweers:
Closing due to no response. Can revisit if you can circle back

Issue #304, opened 2018-12-04
Add support for RxKotlin `subscribeBy`


shaishavgandhi05:
`subscribeBy` should be supported by the lint check. 

I guess we don't need error-prone support since `subscribeBy` is Kotlin only and error-prone doesn't work with Kotlin (?).

hzsweers:
I don't think it should be, it's not part of this API (I've seen 15 different subscribeBy alternatives). We could maybe open up configuration to add more subscribe methods instead

shaishavgandhi05:
Just realized that we don't support `subscribeBy` method in the `ObservableSubscribeProxy` so that renders the issue moot. 

The configuration stuff is interesting but I fear it would be used by very few, while overcomplicating (and probably slowing down) our lint checks.

hzsweers:
Alright, let's close this for now

PR #439, opened 2021-01-15
Add lint check support for kotlin extensions 

Rajin9601:
**Description**:

Lint check should be able to check RxKotlin's subscribeBy.
Looking at the discussion in #304, I get that the library doesn't want to add support only for RxKotlin because it's not part of the api.

So this PR supports kotlin extensions with configuration in gradle.properties. 

For the concern that it would slow down lint-check, I couldn't check for big project but I think this implementation would not slow down lint check much for those who don't use the added configuration. 

**Related issue(s)**: #304

### DIFF
--- a/docs/lint-check.md
+++ b/docs/lint-check.md
@@ -61,6 +61,14 @@ autodispose.lenient=true
 
 The default value of this is `false`. 
 
+### Kotlin Extension
+
+By default, subscribe and subscribeWith methods are checked. To support other subscribe methods such as subscribeBy in RxKotlin, you can add your own subscribe extensions.
+In your **app-level** `gradle.properties` files, add kotlin extension functions in format of `{full package name for extension's scope}#{functionName}` and comma-separated-values like so:
+```groovy
+autodispose.kotlinExtensionFunctions="io.reactivex.rxjava3.kotlin.subscribers#subscribeBy,com.sample.app.SubscribeExt#subscribe2"
+```
+
 #### Examples
 ```java
 // This is allowed in lenient mode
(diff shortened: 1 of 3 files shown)


docs/lint-check.md
```diff
@@ -61,6 +61,14 @@ autodispose.lenient=true
 
 The default value of this is `false`. 
 
+### Kotlin Extension
+
+By default, subscribe and subscribeWith methods are checked. To support other subscribe methods such as subscribeBy in RxKotlin, you can add your own subscribe extensions.
+In your **app-level** `gradle.properties` files, add kotlin extension functions in format of `{full package name for extension's scope}#{functionName}` and comma-separated-values like so:
+```groovy
```
ZacSweers, line 68:
```suggestion
In your **app-level** `gradle.properties` files, add kotlin extension functions in format of `{full package name for extension's scope}#{functionName}` and comma-separated-values like so:

```groovy
```

docs/lint-check.md
```diff
@@ -61,6 +61,14 @@ autodispose.lenient=true
 
 The default value of this is `false`. 
 
+### Kotlin Extension
+
+By default, subscribe and subscribeWith methods are checked. To support other subscribe methods such as subscribeBy in RxKotlin, you can add your own subscribe extensions.
```
ZacSweers, line 66:
```suggestion
By default, `subscribe` and `subscribeWith` methods are checked. To support other subscribe methods such as `subscribeBy` in RxKotlin, you can add your own subscribe extensions.
```

ZacSweers (commented):
Seems ok to me, apologies for the delay. @ShaishavGandhi what do you think?

ShaishavGandhi (approved):
LGTM! Let me take a look at the build issues and then we can merge it.

Rajin9601:
I updated this branch with `uber:main` and passed the CI tests!
It seems like this PR can be merged now.

ZacSweers:
Thanks!

PR #443, opened 2021-04-08
Remove jcenter() repository from project

rafaeltoledo:
Change to avoid things to break after [JCenter sunset](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/).


ZacSweers (approved):
Thanks!

PR #448, opened 2021-07-18
Update annotated packages for NullAway

msridhar:
**Description**: Update NullAway's annotated packages setting to point to the new source package, `autodispose2`.

It's possible that `io.reactivex.rxjava3` should also be treated as an annotated package (if you want to prevent passing of `null` into Rx APIs), but I'm not sure.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -197,7 +197,7 @@ subprojects {
       project.tasks.withType(JavaCompile).configureEach {
         options.errorprone.nullaway {
           severity = CheckSeverity.ERROR
-          annotatedPackages.add("com.uber")
+          annotatedPackages.add("autodispose2")
         }
       }
       project.tasks.withType(Test) {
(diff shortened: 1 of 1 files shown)


ZacSweers (approved):
Nice catch. Yeah we would want to treat RxJava this way too, but I can follow up with that change separately if you want.

msridhar:
> Nice catch. Yeah we would want to treat RxJava this way too, but I can follow up with that change separately if you want.

If you have time to take a look that would be good.  I gave it a quick try but got errors in `HalfSerializer`:

```
HalfSerializer.java:82: error: [NullAway] passing @Nullable parameter 'error.terminate()' where @NonNull is required
        subscriber.onError(error.terminate());
                                          ^
HalfSerializer.java:148: error: [NullAway] passing @Nullable parameter 'error.terminate()' where @NonNull is required
        observer.onError(error.terminate());
```

The code looks safe to me but not sure exactly how you want to suppress these.

ZacSweers:
yeah we'll need to take a larger pass at re-enabling some parts of it

